PR #9, opened 2015-11-19
fix compiler error due to missing support of snprintf on VC11

yidongVSI:
snprintf is not included in Visual Studio 2012, 2013


dstoup:
+1

hjmjohnson:
+1

PR #14, opened 2015-11-23
Remove attempt at forcing try_compiles to re-run as it broke things.

billhoffman:
The CMake code here tried to get cmake to re-run try compile code when things
were changed. As a side effect it kept all the trycompiles from running
and set most of them to true all the time which caused things not to
build.


dstoup:
Branch looks good.
For the record, here is a dashboard which shows the patch working. The Nightly is with DCMTK included without the branch and the experimental has the branch merged.

https://open.cdash.org/index.php?project=vxl&date=20151124

PR #18, opened 2015-11-27
Restrict python versions. BRL python calls are not 3.x compliant.

dstoup:
On a system with python 3.x install, the BRL python calls will fail to
complile. Set the supported version numbers in the FindPython module.

The *.kitware experimental dashboards include this branch and show it working


dstoup:
@hjmjohnson ... No, I don't think we do at all. The intent of this branch was just to block 3.x support to fix compiler issues.

I dislike this FindModule in general I was planning to rework it a bit and remove older python support in the process. I wanted to do that task separately after your CMake update lands. It will be less urgent then.

hjmjohnson:
:+1:   GREAT!  Merge Away!

dstoup:
Thanks for the input!

PR #20, opened 2015-11-30
Fix link errors

dstoup:
Fix a variety of link errors on Linux and Mac that arise when building shared libraries.


hjmjohnson:
@dstoup This looks great!

PR #39, opened 2015-12-07
Modifications to enable use of gcc5.

jhoare:
Added references to GCC5 in tests, as well as in preprocessor macros.


hjmjohnson:
@dstoup This looks like a reasonable addition.

dstoup:
@hjmjohnson I agree. I probably would have merged already, but don't have a gcc 5 machine to verify the build. I does look reasonable though so I am inclined to merge this tonight before leaving ( and prepare to scramble in the morning in case all goes haywire :)

PR #46, opened 2015-12-08
Breg3d ekf camera optimizer test fix

decrispell:
Fix some tests failing due to SEGFAULT on MacOS


decrispell:
Note that the last commit (87c46de) is the one that actually fixed the segfault. The others are cleanup.

dstoup:
Thanks @decrispell, does this patch fix a dashboard test, or just something you've observed?

dstoup:
... sorry, I see my answer in the title :(

decrispell:
Yeah, I originally was just trying to fix the test referenced in the title, but the fix actually seems to have fixed many of the segfault failures seen here:  https://open.cdash.org/viewTest.php?onlyfailed&buildid=4137668

dstoup:
This is excellent, thank you! It will be great to see some green on the dashboards. I cycled it through my mac dashboards just fine and it knocked out half of the issues, ready to go.

PR #48, opened 2015-12-09
test variable, not evaluated variable in CMakeLists.txt

decrispell:
Fixed what seems to be a bug in the brl/bbas/bvgl/algo  CMakeLists.txt, which was preventing an auxiliary file from being copied to the library directory.  This in turn caused the "sdet_test_sel" test to crash.

The crash is fixed, but now there are some failing tests that were not previously running.


dstoup:
Thanks!

PR #51, opened 2015-12-10
Cast fabs input to double to disambiguate call for VS.

dstoup:
This patch fixes the compiler issues on Windows in, what seems like, the most maintenance-free way, but I'm certainly open to a better approach.


decrispell:
We will look into a way to remove the vgl_plane_3d+int- instantiation (again) dependency since it will probably give unexpected results if actually used, but for now this looks good to me.

dstoup:
Thanks for your comments @decrispell 
I'll merge now so we can get closer to a good Windows build. I will write up the vgl_plane_3d+int- instantiation issue in the tracker so we don't lose it. Feel free to edit my description if I don't get it quite right.

PR #56, opened 2015-12-11
Vgl infinite line 3d fixes

decrispell:
These commits should fix the sporadically failing "boxm_util_test_plane_ransac" test.  The problem was rooted in the vgl_infinite_line_3d class.


decrispell:
Thinking about this further, I think the proposed change actually just moves the singularity in the vgl_infinite_line_3d representation from one direction (+x) to another (-z) which isn't unlucky enough to be as well-tested (indirectly by  boxm_util_test_plane_ransac).  Closing until I can verify or fix.

PR #62, opened 2015-12-16
Increase default template-depth.

dstoup:
When CMAKE_CXX_STANDARD is set to 11, we have issues with the current template-depth values set for 2 files. Increasing the depth to 50 fixes the compiler errors.


dstoup:
To prove the need for and functionality of this branch , I've set up a machine to run cxx11 builds. Linux only at the moment, but I'll expand that shortly. Today's dashboard shows ceres failing in the nightly and passing in the experimental. The two branches are identical with the exception of this patch.

https://open.cdash.org/index.php?project=vxl

PR #87, opened 2016-01-06
ENH: Add methods to flatten vnl_matrix_fixed

DVigneault:
Unit tests are provided.  The `vnl_matrix` implementation is also cleaned slightly (should have fewer cache misses).


DVigneault:
@hjmjohnson that's a good point, thank you very much for reviewing!  This patch copies in the data directly, but does so row-wise.  (P.S. It looks like I accidentally removed your comments when I amended the commit--sorry about that, I still have lots to learn about `git`.)

hjmjohnson:
@dstoup @DVigneault This patch looks good to merge now.

PR #98, opened 2016-01-12
Ffmpeg 2.8.4

yanggehua:
These are changes to work with FFMpeg 2.8.4 (recent release).  Please review. 

This is the first time I generated a pull request for the new VXL repo.  Do not hesitate to let me know if there is problem.


core/vidl/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ IF( FFMPEG_FOUND )
                                      vidl_ffmpeg_ostream_v4.txx
       vidl_ffmpeg_istream_v0.9.txx   vidl_ffmpeg_ostream_v0.9.txx
                                      vidl_ffmpeg_ostream_v0.11.txx
+      vidl_ffmpeg_istream_v56.txx    vidl_ffmpeg_ostream_v56.txx
```
mleotta: The naming convention for these files is based on the ffmpeg major/minor version rather than the libavcodec version number.  So these files should be named vidl_ffmpeg_*stream_v2.8.txt.  You can see that there are lots of gaps in supported versions here.  We haven't updated this since FFMPEG v0.11!

The exception here (which is confusing) are the files numbered v1, v2, v3, v4.  This was an arbitrary numbering we used before FFMPEG was making regular versioned releases.  Long ago the FFMPEG version rarely changed (I think it was v0.5), but they keep making changes to the API and Linux distros would pick up those and release them.  So the original file naming was arbitrary.  At this point we should probably either throw away those original version or renamed them in a better way (maybe v1beta.txt or something).

The reason not to use the libavcodec version number here is that we wanted to make it more clear which versions of FFMPEG were supported by looking at the files.  Also we are checking both libavformat and libavcodec versions in different parts of the code and the relationship between each of these and FFMPEG version is not clear.  That said, maybe it would be cleaner to just use libavcodec and libavformat versions to name the ostream and istream files respectively.

core/vidl/vidl_ffmpeg_convert.cxx
```diff
@@ -27,9 +28,9 @@ extern "C" {
 }
 #endif
 
-
 //--------------------------------------------------------------------------------
 
+#if LIBAVCODEC_VERSION_MAJOR < 56
```
mleotta: There is a large block of code here that has been duplicated and modified for different versions.  It's hard to see what has changed, but it looks to be mostly the same.  Can you put the #if around just the parts that have changed to avoid so much code duplication?

core/vidl/vidl_ffmpeg_convert.h
```diff
@@ -46,5 +47,26 @@ vidl_pixel_format_from_ffmpeg(PixelFormat ffmpeg_pix_fmt);
...
+// to \p out_frame->pixel_format() and stored in \p out_frame->data()
+// If the output frame data is not the correct size new memory
+// will be allocated
+bool vidl_ffmpeg_convert(const vidl_frame_sptr& in_frame,
+                               vidl_frame_sptr& out_frame);
+
+
+//: Find the vidl pixel format that matches a FFMPEG one
+vidl_pixel_format
+vidl_pixel_format_from_ffmpeg(AVPixelFormat ffmpeg_pix_fmt);
```
mleotta: Ah, now I see that it's the AVPixelFormat type that has changed.  Can we typedef or define PixelFormat to AVPixelFormat for older versions to avoid the code duplication?

core/vidl/vidl_ffmpeg_ostream.cxx
```diff
@@ -30,8 +30,10 @@ extern "C" {
 # include "vidl_ffmpeg_ostream_v4.txx"
 #elif LIBAVCODEC_BUILD < ((54<<16)+(23<<8)+100)  // before ver 54.23.100
 # include "vidl_ffmpeg_ostream_v0.9.txx"
-#else
+#elif LIBAVCODEC_BUILD < ((56<<16)+(0<<8)+100)  // before ver 56.0.100
```
mleotta: Ideally what we want here is the last version that is supported by the vidl_ffmpeg_ostream_v0.11.txx implementation.  However, since there is a huge range of version that we don't support, it's broken either way for those in between versions.  So it's not critical that this is addressed in this branch. 

A better approach in the future would be to check for those unsupported version and throw an error.  I think Kitware has code to do that that we will contribute in the future.

core/vidl/vidl_ffmpeg_ostream_v56.txx
```diff
@@ -0,0 +1,597 @@
+// This is core/vidl/vidl_ffmpeg_ostream_v0.11.txx
```
mleotta: You need to update all this header info and include guards

core/vidl/vidl_ffmpeg_istream_v56.txx
```diff
@@ -0,0 +1,590 @@
+// This is core/vidl/vidl_ffmpeg_istream_v0.9.txx
```
mleotta: You need to update the header info and the include guards

core/vidl/vidl_ffmpeg_istream_v0.9.txx
```diff
@@ -474,7 +474,7 @@ vidl_ffmpeg_istream::current_frame()
 
     // If the pixel format is not recognized by vidl then convert the data into RGB_24
     vidl_pixel_format fmt = vidl_pixel_format_from_ffmpeg(enc->pix_fmt);
-    if (fmt == VIDL_PIXEL_FORMAT_UNKNOWN)
+    if (fmt == VIDL_PIXEL_FORMAT_UNKNOWN || fmt == VIDL_PIXEL_FORMAT_YUV_420P)
```
mleotta: what is the purpose of this change?

yanggehua: This was an old change that I kept to myself.  With the little recollection that I have, I believe I ran into a some problem when decoding a raw video stream that was encoded in YUV_420P format.  After adding this or condition, I was able to get the full RGB image decoded.   I had to admit that the code in this way reads a little odd though.

mleotta: The purpose of this block was to catch FFMPEG pixel formats that vidl does not know about or cannot represent.  This is a fallback option which converts the image into a default format that vidl does understand (RGB_24).  Otherwise the memory is wrapped (shallow copy) in a vidl_frame with no conversion.  The purpose of this approach is to postpone format conversion until absolutely needed.  By adding YUV_420P to this list you are forcing vidl to convert it early.  If, for example, someone wanted to write that frame back out into a new video as YUV_420P, no conversion would be needed at all, but by adding this change every frame would be converted to RGB_24 and then back again.

If you were having problems it's probably due to a bug elsewhere in converting from YUV_420P.  This is not the appropriate fix.

yanggehua: Your reasoning is solid.  I will remove it from the branch.

core/vidl/vidl_ffmpeg_convert.cxx
yanggehua: Used typedef and macro definition to map the old types and values to the new ones.  Keeping the new ones as untouched, assuming this is the direction to go.   Code is pushed.

core/vidl/vidl_ffmpeg_ostream_v56.txx
yanggehua: Fixed and pushed.

core/vidl/vidl_ffmpeg_istream_v56.txx
yanggehua: Fixed and pushed.

PR #108, opened 2016-01-14
replace checks for "linux" preprocessor definition with "__linux__"

decrispell:
The "linux" preprocessor definition is nonconforming and is suppressed
by gcc when supplied with flags such as "-std=X" and "-pedantic".
The double underscore equivalent is supported at least as far back as
gcc v3.1.1.

see:
https://gcc.gnu.org/onlinedocs/gcc-3.1.1/cpp/System-specific-Predefined-Macros.html#System-specific%20Predefined%20Macros
http://sourceforge.net/p/predef/wiki/OperatingSystems/


hjmjohnson:
@decrispell Looks Great!

@dstoup Ready to merge.

Issue #107, opened 2016-01-14
Linking error when using the head of Master branch


yanggehua:
I encountered the following linking error with MSVC 2015.  It is missing instantiation file for vnl_sym_matrix< vcl_complex<float> >, as needed by commit c68bf93aecaf79ebfa986dd394c7869016534d7a.  

10>test_complexify.obj : error LNK2019: unresolved external symbol "public: class vnl_sym_matrix<class std::complex<float> > & __cdecl vnl_sym_matrix<class std::complex<float> >::update(class vnl_sym_matrix<class std::complex<float> > const &,unsigned int)" (?update@?$vnl_sym_matrix@V?$complex@M@std@@@@QEAAAEAV1@AEBV1@I@Z) referenced in function "public: __cdecl vnl_sym_matrix<class std::complex<float> >::vnl_sym_matrix<class std::complex<float> >(class vnl_sym_matrix<class std::complex<float> > const &)" (??0?$vnl_sym_matrix@V?$complex@M@std@@@@QEAA@AEBV0@@Z)
10>D:\Code\vxl\bin-vc14-x64\core\vnl\tests\Debug\vnl_test_all.exe : fatal error LNK1120: 1 unresolved externals

DVigneault:
@yanggehua Adding the instantiation now--sorry for adding a bug.  Does this need to be done for `vnl_diag_matrix` also?  Which files are you including to get the error?

dstoup:
I believe this issue is resolved with merged branch.

PR #109, opened 2016-01-15
WIP: Overload vnl_complexify for additional containers.

DVigneault:
@yanggehua brought up a linker error that he was experiencing [1], having to do with template instantiation.  I wasn't able to reproduce this error, but I was able to reproduce a similar one (see the commented out unit test, which complains about undefined reference to `operator=`).  I think this may have to do with the fact that `VNL_SYM_MATRIX_INSTANTIATE` [2] does not instantiate it, but I wasn't able to figure out the syntax for adding it.

@thewtex @hjmjohnson I'm not very familiar with vxl's system for instantiating templates--any guidance is appreciated!

[1] https://github.com/vxl/vxl/issues/107
[2] https://github.com/vxl/vxl/blob/master/core/vnl/vnl_sym_matrix.txx

### DIFF
--- a/core/vnl/tests/test_complexify.cxx
+++ b/core/vnl/tests/test_complexify.cxx
@@ -21,10 +21,12 @@ void test_complexify()
 
   typedef float TReal;
   const unsigned int length = 2;
-  const TReal value = 7.5;
+  const TReal r = 7.5;
+  const TReal i = 8.5;
 
   vcl_cout << "Testing vnl_vector" << vcl_endl;
-  vnl_vector<TReal> r_vector(length,value);
+  const vnl_vector<TReal> r_vector(length,r);
+  const vnl_vector<TReal> i_vector(length,i);
   vnl_vector<vcl_complex<TReal> > c_vector
     = vnl_complexify(r_vector);
   for (unsigned int i = 0; i < length; ++i)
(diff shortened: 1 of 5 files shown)


core/vnl/tests/test_complexify.cxx
```diff
@@ -101,6 +157,16 @@ void test_complexify()
       TEST("vnl_sym_matrix imag",0.0,c_sym_matrix(r,c).imag());
       }
     }
+//  c_sym_matrix
```
DVigneault: If this is uncommented, it causes a linker error:

    CMakeFiles/vnl_test_all.dir/test_complexify.cxx.o: In function `test_complexify()':
    test_complexify.cxx:(.text+0x11c7): undefined reference to `vnl_sym_matrix<std::complex<float> >::operator=(vnl_sym_matrix<std::complex<float> > const&)'
    collect2: error: ld returned 1 exit status
    make[2]: *** [core/vnl/tests/vnl_test_all] Error 1
    make[1]: *** [core/vnl/tests/CMakeFiles/vnl_test_all.dir/all] Error 2
    make: *** [all] Error 2

thewtex:
@DVigneault How about creating the analog to core/vnl/Templates/vnl_sym_matrix+double-.cxx for the complex float and complex double types?

DVigneault:
Thanks @thewtex, that does it for me (with a clean rebuild).  I hope that this will solve the linker errors for @yanggehua also.

PR #156, opened 2016-02-05
optionally build without ffmpeg support

decrispell:
Since supporting whatever version of ffmpeg happens to be available seems to be a tricky issue, I added a "WITH_FFMPEG" cmake option that will not bother searching for it if turned off.  The option is on by default so default behavior shouldn't change.   @mleotta mentioned in pull request #98  that there may be support for ffmpeg version checking added in the future, so this option could probably be removed at that point.

### DIFF
--- a/contrib/brl/bseg/boxm2/class/CMakeLists.txt
+++ b/contrib/brl/bseg/boxm2/class/CMakeLists.txt
@@ -10,7 +10,7 @@ set(boxm2_class_sources
    )
 
 add_library(boxm2_class  ${boxm2_class_sources})
-target_link_libraries(boxm2_class boxm2 bvgl vpgl vnl vidl vil_algo vil vgl_algo vgl_xio vgl vbl vul vsl vcl)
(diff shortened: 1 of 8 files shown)


mleotta:
@decrispell, should this be a CMake option instead of just a variable?  For the record, you can disable FFMPEG now by manually setting the FFMPEG variables in CMake to `IGNORE`.  However, that is not well known and not user friendly.  Your approach makes it a little easier, but I assume you need to set the option on the command line to use it.  I think if it was a CMake option you could set it either on the command line or in the CMake configuration GUIs.

decrispell:
@mleotta It's a cache variable, so it can be set it from the GUI.  To be honest I'm not sure what the difference between a variable declared with option vs. set with "cache" is.  Is there a reason to prefer option?

mleotta:
hmm... other than `option` being more explicit and potentially easier to follow when browsing the code, I'm not sure.  I need to defer to the experts.  Maybe @bradking or @billhoffman can comment.

hjmjohnson:
I agree that option is better documented and easier to maintain methods for specifying optional features in cmake.

decrispell:
Done.

PR #157, opened 2016-02-05
COMP: Fixes Value stored but never read warnings

ajjl:
In many of the files converted from FORTRAN, values are stored
to variables and then never read. Ex:
vxl/v3p/netlib/tests/lsqr-test.c|715 col 5| warning: Value stored to 'damp4' is never read
||     damp4 = (float)1e-4;
||     ^       ~~~~~~~~~~~
In such cases the line is commented out.
@hjmjohnson This fixes the stored but not read warnings revealed by the clang static analyzer


hjmjohnson:
@ajjl  Failing test is known to be unstable across different runs.

PR #172, opened 2016-02-10

ajjl:
warnings illitsrate faulty logic, and appear when branches in the code

### DIFF
--- a/contrib/brl/bpro/core/vpgl_pro/processes/vpgl_get_bounding_box_process.cxx
+++ b/contrib/brl/bpro/core/vpgl_pro/processes/vpgl_get_bounding_box_process.cxx
@@ -88,10 +88,10 @@ bool vpgl_get_bounding_box_process(bprb_func_process& pro)
 
     //intersect each ray with z plane
     vgl_point_3d<double> ulp, urp, blp, brp;
-    bool good =    vgl_intersection(ul, zp, ulp);
-    good = good && vgl_intersection(ur, zp, urp);
-    good = good && vgl_intersection(bl, zp, blp);
-    good = good && vgl_intersection(br, zp, brp);
+    vgl_intersection(ul, zp, ulp);
+    vgl_intersection(ur, zp, urp);
+    vgl_intersection(bl, zp, blp);
+    vgl_intersection(br, zp, brp);
 
     //convert the four corners into image coordinates
     typedef vgl_polygon<double>::point_t        Point_type;
--- a/contrib/brl/bseg/boxm2/ocl/algo/boxm2_ocl_paint_online_color.cxx
+++ b/contrib/brl/bseg/boxm2/ocl/algo/boxm2_ocl_paint_online_color.cxx
@@ -364,14 +364,13 @@ bool boxm2_ocl_paint_online_color::paint_scene_with_weights(boxm2_scene_sptr    
       status = clFinish(queue);
       check_val(status, MEM_FAILURE, "BEFORE UPDATE_KERN: clFinish FAILED: " + error_to_string(status));
 
-      bool arg_status = true;
-      arg_status &= update_kern->set_arg( blk_info );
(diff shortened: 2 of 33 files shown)


hjmjohnson:
@decrispell @DVigneault 

Alexander @ajjl is working on code cleanups.  It would be beneficial for his progress if this patch were reviewed expediently so that he can efficiently address the next set of compiler warnings.

contrib/brl/bpro/core/vpgl_pro/processes/vpgl_get_bounding_box_process.cxx
```diff
@@ -88,10 +88,10 @@ bool vpgl_get_bounding_box_process(bprb_func_process& pro)
...
     //intersect each ray with z plane
     vgl_point_3d<double> ulp, urp, blp, brp;
-    bool good =    vgl_intersection(ul, zp, ulp);
-    good = good && vgl_intersection(ur, zp, urp);
-    good = good && vgl_intersection(bl, zp, blp);
-    good = good && vgl_intersection(br, zp, brp);
+    vgl_intersection(ul, zp, ulp);
+    vgl_intersection(ur, zp, urp);
+    vgl_intersection(bl, zp, blp);
+    vgl_intersection(br, zp, brp);
```
decrispell: Wouldn't it be better to keep the return values from vgl_intersection, but add code to actually check it and report an error if it fails?

hjmjohnson: David,

That is a good point,   but there is a limit to the  amount of effort we can put into this.  We have expertise in removing problematic code that has been there for a long time, but not in the complete code base.

To be clear, our goal is to remove these “noisy” warnings so that we can more effectively treat the “real warnings”  There are hundreds of other warnings are are more indicative of bugs or other logical errors.

I will ask @ajjl to revert this change and improve the test, but let’s try to find a balance between “optimal code changes” and “evolutionary code changes”.

========

Thank you for your quick response.
--

dstoup:
I agree with Daniel on this one. The warnings are there because we're doing something wrong. If we simply remove the code, the warning goes away, but the problem persists. I agree also with Hans that we should strike a balance, but in this case, the code expresses the original intent and it's only a line of code to warn when !good. That seems like a pretty decent balance to me.

hjmjohnson:
Thanks all!  

@Ajjl  Fix and resubmit.  
--

hjmjohnson:
@ajjl Review entire patch set and determine if there are other instances in tests where you could improve the test by using the value rather than removing it.

contrib/brl/bseg/boxm2/ocl/algo/boxm2_ocl_paint_online_color.cxx
```diff
@@ -364,14 +364,13 @@ bool boxm2_ocl_paint_online_color::paint_scene_with_weights(boxm2_scene_sptr
...
-      arg_status &= update_kern->set_arg( aux1 );
-      arg_status &= update_kern->set_arg( aux2 );
-      arg_status &= update_kern->set_arg( cl_output.ptr() );
+      update_kern->set_arg( blk_info );
+      update_kern->set_arg( mog );
+      update_kern->set_arg( num_obs );
+      update_kern->set_arg( aux0 );
+      update_kern->set_arg( aux1 );
+      update_kern->set_arg( aux2 );
+      update_kern->set_arg( cl_output.ptr() );
```
dstoup: I suspect this block is another which could test and warn rather than remove

yidongVSI:
@ajjl FYI, I have reviewed all the 'volm' associated parts and they are all fine

hjmjohnson:
Thanks Yi.

--

ajjl:
Thanks for the reviews everyone! I'll redo the patch.

PR #174, opened 2016-02-11

ajjl:
warnings illusrate faulty logic, and appear when branches in the code


ajjl:
@hjmjohnson @decrispell @DVigneault 
Thanks for the earlier feedback! Here is the re-done patch, I hope this address your concerns.

decrispell:
Looks good to me, Thanks for the changes!  One minor style point, I think it's more readable to just check bool values directly `if (good)` , rather than the technically redundant `if (good == true)`.  It's probably debatable, though.

ajjl:
I can fix those.

PR #175, opened 2016-02-12

ajjl:
warnings illusrate faulty logic, and appear when branches in the code


ajjl:
@hjmjohnson @decrispell  @DVigneault 
Let me know what you think.

PR #178, opened 2016-02-13
Orbit exec augmentations

deuscovrigus:
More improvements to orbit display exec


hjmjohnson:
@deuscovrigus Can you please confirm that 
The following tests FAILED:
	692 - bsta_algo_test_fit_gaussian (Failed)

is not related to the changes in this pull request?  I think it is a failure (perhaps due to use of unfortunate random number generator selection, or unitialized variables).

I'll merge if you can confirm that this failure is not related to this patch set.

deuscovrigus:
This test is not associated with the pull request, and i thing the RNG is
at fault. When I'll have time I can take a look at why it's failing.

PR #181, opened 2016-02-17
COMP: Removes "Division by zero" Warnings

ajjl:
including "Division by zero". These warnings arise when
variables are not checked to see if they are zero before being used
as divisors. Most of these warnings would never actually result in
division by zero in real world usage.

### DIFF
--- a/contrib/brl/bseg/bapl/bapl_keypoint_extractor.cxx
+++ b/contrib/brl/bseg/bapl/bapl_keypoint_extractor.cxx
@@ -373,13 +373,18 @@ bapl_lowe_orientation::orient_at( float x, float y, float scale,
   for (unsigned int i=0; i<peaks.size(); ++i) {
     if (histogram[ peaks[i] ] > max) {
       //parabolic interpolation
-      float ypos = histogram[ (peaks[i]+1)%num_bins_ ];
-      float yneg = histogram[ (peaks[i]-1)%num_bins_ ];
-      float dy   = (ypos - yneg)/2.0f;
-      float d2y  = 2.0f*histogram[ peaks[i] ] - ypos - yneg;
-      float dx = 6.28319f/num_bins_;
-      float angle = (float(peaks[i])+dy/d2y)*dx;
-      orientations.push_back(angle);
+      if(num_bins_ != 0) {
+        float ypos = histogram[ (peaks[i]+1)%num_bins_ ];
+        float yneg = histogram[ (peaks[i]-1)%num_bins_ ];
+        float dy   = (ypos - yneg)/2.0f;
+        float d2y  = 2.0f*histogram[ peaks[i] ] - ypos - yneg;
+        float dx = 6.28319f/num_bins_;
+        float angle = (float(peaks[i])+dy/d2y)*dx;
+        orientations.push_back(angle);
+      }
+      else {
+        vcl_cout << "ERROR: Division by 0" << vcl_endl;
+      }
     }
   }
 }
--- a/contrib/gel/vmal/vmal_dense_matching.cxx
+++ b/contrib/gel/vmal/vmal_dense_matching.cxx
@@ -336,7 +336,11 @@ void vmal_dense_matching::disparity_map(vmal_multi_view_data_edge_sptr mvd_edge,
   {
     for (int j=0; j<w; j++)
     {
-      int value=vmal_round_int((map(j,i)-min_disparity)*255/max_disparity);
+      int value = 0;
+      if(max_disparity != 0 )
+      {
+        value=vmal_round_int((map(j,i)-min_disparity)*255/max_disparity);
+      }
       buf[i*w+j]=(unsigned char)value;
     }
   }
--- a/core/vul/vul_psfile.cxx
+++ b/core/vul/vul_psfile.cxx
@@ -287,7 +287,10 @@ void vul_psfile::print_greyscale_image(unsigned char* buffer, int sizex, int siz
               index += int(*(buffer + (pixel_number+m+n*width)));
               ++number_of_pixels_sampled;
             }
-        index/=number_of_pixels_sampled; // Average the pixel intensity value.
+        if(number_of_pixels_sampled != 0)
+        {
+          index/=number_of_pixels_sampled; // Average the pixel intensity value.
+        }
       }
 
       // write hex pixel value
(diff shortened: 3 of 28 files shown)


hjmjohnson:
@DVigneault  Could you please review?

contrib/gel/vmal/vmal_dense_matching.cxx
```diff
@@ -336,7 +336,11 @@ void vmal_dense_matching::disparity_map(vmal_multi_view_data_edge_sptr mvd_edge,
   {
     for (int j=0; j<w; j++)
     {
-      int value=vmal_round_int((map(j,i)-min_disparity)*255/max_disparity);
+      int value = 0;
+      if(max_disparity != 0 )
+      {
+        value=vmal_round_int((map(j,i)-min_disparity)*255/max_disparity);
+      }
```
dstoup: Should there be a warning or error logged here?

core/vul/vul_psfile.cxx
```diff
@@ -287,7 +287,10 @@ void vul_psfile::print_greyscale_image(unsigned char* buffer, int sizex, int siz
               index += int(*(buffer + (pixel_number+m+n*width)));
               ++number_of_pixels_sampled;
             }
-        index/=number_of_pixels_sampled; // Average the pixel intensity value.
+        if(number_of_pixels_sampled != 0)
+        {
+          index/=number_of_pixels_sampled; // Average the pixel intensity value.
+        }
```
dstoup: Log warning or error since computation is skipped.

core/vul/vul_psfile.cxx
```diff
@@ -443,7 +446,10 @@ void vul_psfile::print_color_image(unsigned char* data, int sizex, int sizey)
                 index += int(*(data+(pixel_number+(m+n*sizex)*bytes_per_pixel)));
                 ++number_of_pixels_sampled;
               }
-          index/=number_of_pixels_sampled;  // average the pixel intensity
+          if(number_of_pixels_sampled != 0)
+          {
+            index/=number_of_pixels_sampled;  // average the pixel intensity
+          }
```
dstoup: same ....

dstoup:
Looks good except for the few places where computation is skipped because of 0 value, but user is never notified. Could lead to unpredictable results.

contrib/brl/bseg/bapl/bapl_keypoint_extractor.cxx
```diff
@@ -373,13 +373,18 @@ bapl_lowe_orientation::orient_at( float x, float y, float scale,
...
+        float ypos = histogram[ (peaks[i]+1)%num_bins_ ];
+        float yneg = histogram[ (peaks[i]-1)%num_bins_ ];
+        float dy   = (ypos - yneg)/2.0f;
+        float d2y  = 2.0f*histogram[ peaks[i] ] - ypos - yneg;
+        float dx = 6.28319f/num_bins_;
+        float angle = (float(peaks[i])+dy/d2y)*dx;
+        orientations.push_back(angle);
+      }
+      else {
+        vcl_cout << "ERROR: Division by 0" << vcl_endl;
```
DVigneault: One thought--should execution stop entirely here?  It might be problematic if it continues while just printing out an error.  I don't know if these are used much in the library, but maybe this would constitute a "domain error."  https://github.com/vxl/vxl/blob/647386048f3f71ed6d9e7d67de5d11dab5ddacfc/vcl/generic/vcl_stdexcept.h

DVigneault:
Besides @dstoup's comments, my only other question is whether it is more appropriate to return, as you have done, or to raise an exception.  @dstoup and @hjmjohnson will be more qualified to say which makes more sense.

hjmjohnson:
I’m not sure what “the most appropriate” is.  This is not our code, and actually we don’t even use it.

We are trying to clean up compiler identified warnings, but in some cases we are making our best guess with what to do.

I think that even an assert might be appropriate or exception throwing.    I’m sure the current code is wrong if the denominator is zero.  The path of least surprise (explicitly have the program stop with exception or assert) in these bad cases is probably the best choice.  Then if this causes a problem for someone, they can easily identify the source location of the problem and make appropriate actions.

Hans

dstoup:
My only beef with asserts is that when it fails on production code only,
I'm still left with days of really difficult debugging.

I don't understand the code well enough to know the implications of
continued processing vs. exit with failure. I personally like exceptions
assuming they're going to be caught somewhere, though I'm not always in the
majority on that opinion.


On Thu, Feb 18, 2016 at 12:37 PM, Hans Johnson <notifications@github.com>
wrote:

> I’m not sure what “the most appropriate” is. This is not our code, and
> actually we don’t even use it.
>
> We are trying to clean up compiler identified warnings, but in some cases
> we are making our best guess with what to do.
>
> I think that even an assert might be appropriate or exception throwing.
> I’m sure the current code is wrong if the denominator is zero. The path of
> least surprise (explicitly have the program stop with exception or assert)
> in these bad cases is probably the best choice. Then if this causes a
> problem for someone, they can easily identify the source location of the
> problem and make appropriate actions.
>
> Hans
>
>
> From: Davis Vigneault <notifications@github.com>
> Reply-To: vxl/vxl <reply@reply.github.com>
> Date: Thursday, February 18, 2016 at 11:30 AM
> To: vxl/vxl <vxl@noreply.github.com>
> Cc: Hans Johnson <hans.j.johnson@gmail.com>
> Subject: Re: [vxl] COMP: Removes "Division by zero" Warnings (#181)
>
>
> Besides @dstoup's comments, my only other question is whether it is more
> appropriate to return, as you have done, or to raise an exception. @dstoup
> and @hjmjohnson will be more qualified to say which makes more sense.
>
> —
> Reply to this email directly or view it on GitHub.
>
>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vxl/vxl/pull/181#issuecomment-185828332>.
>



-- 
David Stoup
R&D Engineer

Kitware, Inc.
28 Corporate Drive
Clifton Park, NY. 12065
518-881-4949 (W)
518-312-3946 (M)
518-371-4573 (F)

ajjl:
Thanks for the comments everyone! I'll look into adding asserts. 
In many places, the code is part of a function that returns true when completed. For example:
contrib/brl/bseg/boxm2/cpp/algo/boxm2_export_oriented_point_clound.cxx
(The second file in the diff)
The documentation is a bit sparse, but to me it looks like the function returns true to indicate that it completes successfully. In this case, doesn't it make sense to just return false, and have the responsibility of checking the return value left to the function caller? Would an assert or exception or something also be necessary?

DVigneault:
@ajjl Yes, I think that for the functions that return `bool` logging an error and returning `false` is best.  Apologies for not being explicit--my comment was directed towards the function with the inline comment because it doesn't return `bool`.  Thanks for the patch--as long as the warnings are gone and there's something to indicate what went wrong I'm happy.  :-)

PR #186, opened 2016-02-19
optimizer for position, orientation, and focal length  + better test coverage

decrispell:
Adds a function to optimize position, orientation and focal  length to vpgl_optimize_camera, as well as some test coverage.  Previously, no tests were actually computed in test_optimize_camera.


hjmjohnson:
@decrispell Thanks for the test that goes along with the new feature.

PR #196, opened 2016-02-22
ENH: Use BUILD_SHARED_LIBS global flag to force

ricortiz:
@hjmjohnson @thewtex I tested this in my Linux box against  [this ITK example](http://www.itk.org/Wiki/ITK/Examples/Math/Trig/Atan2ImageFilter).


thewtex:
LGTM

PR #203, opened 2016-02-26
BUG: Fix cmake errors related to INTERFACE_INCLUDE_DIRECTORY

ricortiz:
property not containing full path in cmake 2.8.11.

@hjmjohnson @andyneff This should fix the cmake errors...


andyneff:
Thanks for this! It turns out that error WAS more important when it came to build systems like ninja, and now I am able to compile again!

Issue #206, opened 2016-02-27
Are VXL exception functions just aborting?


DVigneault:
### TL;DR

According to the [implementation file](https://github.com/vxl/vxl/blob/647386048f3f71ed6d9e7d67de5d11dab5ddacfc/core/vnl/vnl_error.cxx) documentation (there is no documentation in the header), the function `vnl_error_vector_index` should "[r]aise exception for invalid index".  However, my attempts to `throw` and `catch` with this function resulted in program termination.  After drilling down into the preprocessor definitions, it looks as though the function is just calling `std::abort()`, and no exception is being raised.  My questions are:

1.  Am I correct that `vnl_error_vector_index`, and all the other functions in that file, are just calling `std::abort()`?
2.  If so, what should be the actual behavior?  It seems to me that if the implementation is correct, the documentation needs to be updated, or vice versa.

### The Gory Details

Let's say I want to catch an exception thrown in the standard library.  I can do it like this:

```
  std::vector<int> std_vec;
  try {std_vec.at(25);}
  catch (...) { vcl_cout << "This gets caught." << vcl_endl; }
```
I thought perhaps I could do the same for a vxl exception, but the exception does not get caught:
```
  try { vnl_error_vector_index("get", 25); }  // Raise exception
  catch (...) { vcl_cout << "Not caught." << vcl_endl; }
```
After poking around in the code for a while, I found that vxl actually defines its own try/catch structure, but unfortunately that doesn't catch it either:
```
  vcl_try { vnl_error_vector_index("get", 25); }  // Raise exception
  vcl_catch_all { vcl_cout << "Not caught." << vcl_endl; }
```
Both of the vxl examples give an error and terminate execution.
```
vnl_error_vector_index:get: Invalid value 25 specified for index.
```
So I poked around some more and found the implementation of `vxl_error_vector_index` [here](https://github.com/vxl/vxl/blob/647386048f3f71ed6d9e7d67de5d11dab5ddacfc/core/vnl/vnl_error.cxx):

```
//: Raise exception for invalid index
void vnl_error_vector_index (char const* fcn, int index)
{
  //RAISE Error, SYM(vnl_error_vector), SYM(Invalid_Index),
  vcl_cerr << "vnl_error_vector_index:" << fcn
           << ": Invalid value " << index << " specified for index.\n";
  vcl_abort();
}
```
[...]

PR #207, opened 2016-02-27
ENH: Unifying out-of-bounds behavior for vnl containers.

DVigneault:
Currently, the behavior when accessing elements in vnl containers
using the get() and put() functions is inconsistent.  This patch
attempts to unify the behavior somewhat, such that exceptions
are raised when an out-of-bounds index is requested from `get()` or
`put()` and  `VNL_CONFIG_CHECK_BOUNDS` is enabled.

Importantly, the behavior of the functions in `vnl_error.h` has been
modified such that exceptions are raised, rather than simply
aborting.  This makes the behavior of the functions consistent
with their current documentation.

This patch addresses an issue (https://github.com/vxl/vxl/issues/206) raised here.

### DIFF
--- a/core/vnl/tests/test_diag_matrix.cxx
+++ b/core/vnl/tests/test_diag_matrix.cxx
@@ -9,6 +9,7 @@
 #include <vnl/vnl_diag_matrix.h>
 #include <vnl/vnl_matrix_fixed.h>
 #include <vnl/vnl_vector_fixed.h>
+#include <vcl_exception.h>
 
 void test_diag_matrix()
 {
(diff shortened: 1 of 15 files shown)


DVigneault:
> gcc <-- All tests pass
> clang <-- gevd_test_noise fails

All tests pass on my computer (Ubuntu 14.04, g++ 5.3.0).  I see that `gevd_test_noise` has failed in at least one [other pull request](https://github.com/vxl/vxl/pull/145) as well.

DVigneault:
`$ for i in `seq 1 50`; do   ctest -R gevd_test_noise; done`

Using this I'm not able to get it to fail on my machine.

PR #208, opened 2016-02-27
ENH: Syntax highlighting hints for txx files.

DVigneault:
Github doesn't automatically recognize txx files as c++, and therefore does not provide syntax highlighting in the browser.  This patch gives github a "hint" which allows it to provide in-browser syntax highlighting.

### DIFF
--- a/core/vbl/vbl_array_1d.txx
+++ b/core/vbl/vbl_array_1d.txx
@@ -1,4 +1,5 @@
 // This is core/vbl/vbl_array_1d.txx
+// -*- c++ -*-
 #ifndef vbl_array_1d_txx_
 #define vbl_array_1d_txx_
 
--- a/core/vbl/vbl_array_2d.txx
+++ b/core/vbl/vbl_array_2d.txx
@@ -1,4 +1,5 @@
 // This is core/vbl/vbl_array_2d.txx
+// -*- c++ -*-
 #ifndef vbl_array_2d_txx_
 #define vbl_array_2d_txx_
 
--- a/core/vbl/vbl_array_3d.txx
+++ b/core/vbl/vbl_array_3d.txx
@@ -1,4 +1,5 @@
 // This is core/vbl/vbl_array_3d.txx
+// -*- c++ -*-
 #ifndef vbl_array_3d_txx_
 #define vbl_array_3d_txx_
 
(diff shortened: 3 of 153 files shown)


hjmjohnson:
@DVigneault this is a hack that fixes only github.  The problem is more widespread.  We ran into this for ITK (see below).  ITK has git hooks to prevent .txx files from being added in the future.

I would HIGHLY recommend (and emphatically request) that we fix the root problem and change from .txx to .hxx (or .hpp) file extensions.

I was planning to do that before integrating VXL with ITK.  If you could do this I would be forever grateful.

    STYLE: Change for development tool compliance
    
    PREMISE:
    The  use of the .txx file extension for indicating the file
    that is to hold the implementation of a template class is
    only adopted by the ITK community and VNL communities (and has
    slightly “infected” a few other small components of other projects).
    Increasingly tools are being created to help the programmer better
    understand and analyze, and provide hints to proper coding options
    dynamically in the development environment (color coding, auto-completion).
    
    The obscure .txx file requires special configuration of each development
    environment tool to indicate that it is really a c++ header.  For new
    developers this is often a difficult and frustrating task.  This is
    confirmed by the large amount of documentation for just a few of the
    common editors.  Tools that recognize and respect the .hxx extension:
    uncrustify, vim, emacs, VisualStudios, clang, SlickEdit,
    kwrite, etc....)  None of these tools recognize the .txx extension without
    end user customizations.
    
    CONS:
    Many files will be changed, and it will not be backwards compatible for
    those who accidently included the “.txx” files in their projects.  The .txx
    files should not be included in external packages.
    
    BENEFITS:
    Implementation files are properly mapped to C++ language type by default
    for improved development environment experience.
    
    CONSIDERATIONS:
    There is no obvious single correct solution, but in general a common theme
    from other guides is that the final extension should be “.hxx” or
    some suffix that ends with “.h”.  The extension should be all lower case
    letters (to avoid problems on case insensitive file sysetms).
    
    PROPOSAL:
    In ITKv4 (and outside of the ThirdParty directory) change all implementation files
    that currently have a “.txx” extension to have a “.hxx” extension.
    
    An in-depth discussion was held near April 20, 2011 on the Insight-Developers
    mailing list.

hjmjohnson:
@thewtex Matt, do you see any issue with changing from ".txx" to ".hxx" for VXL?

DVigneault:
@hjmjohnson Thanks very much for the in-depth response!  Your proposal sounds like a much better solution.  I will look into this.

hjmjohnson:
@DVigneault This pull request should no longer be needed.

PR #222, opened 2016-03-04
ENH: Adds CMake macro combining  PERFORM_CMAKE_TEST and SET_INVERT

ajjl:
The two macros, PERFORM_CMAKE_TEST and SET_INVERT are often used
together when a false result is expected. This macro combines both
of these together. NOTE: this does not replace the exact function of
the two old macros with the new one. Because PERFORM_CMAKE_TEST skips
its test if the TEST variable is defined, calling the two functions
in succession might have resulted in flipping the desired result, as
SET_INVERT did not know that PERFORM_CMAKE_TEST had decided not to
test. This was assumed to be a bug, and is addressed in the new macro
that combines the two functions.


hjmjohnson:
@ajjl This has been outstanding for so long that it appears to have been abandoned.  Closing for now.  If it needs to be updated an re-implements, the re-open the pull request with a new patch set.

PR #231, opened 2016-03-09
v3p: Fix check for local zlib directory

bradking:
Fix our `if(... AND EXISTS zlib)` condition to check for the source
directory via absolute path.  Otherwise it checks the current working
directory of CMake and may not think the `zlib` source directory exists.


bradking:
Cc: @hjmjohnson

hjmjohnson:
Clang failing test is known to be an unstable test.  Safe to ignore.  bmdl_pro_generate_mesh_process

PR #232, opened 2016-03-10
ENH: Apply across rows and columns of matrix_fixed

DVigneault:
`apply_rowwise` and `apply_columnwise` methods were recently added to `vnl_matrix`.  This patch extends this functionality to `vnl_matrix_fixed`.


DVigneault:
The following failed on clang:

    848 - bmdl_pro_generate_mesh_process (OTHER_FAULT)

I suspect that this is unrelated, but don't know if this test is known failure.

PR #233, opened 2016-03-10
Remove unused `.dsptemplate` files

bradking:
These files were needed only for the CMake VS 6 generator that we
no longer support.


bradking:
Cc: @thewtex

bradking:
Cc: @hjmjohnson

hjmjohnson:
Deleted code is easy to maintain!  Thank you for noticing.

thewtex:
+1 Thanks!

PR #237, opened 2016-03-13
Backwards compatible vcl defs

hjmjohnson:


hjmjohnson:
@thewtex Matt,  This version of VXL should provide better backwards compatibility to allow TubeTK (and other Remote modules) to build without making changes.

Currently those packages do NOT include vcl_* headers (if they did this would not be necessary).  They instead include vnl_* headers which used to also include vcl_* headers.  I've updated VXL to work around this and provide a cleaner solution.

thewtex:
Excellent.

PR #239, opened 2016-03-14
BUG: Error in ordering of #defines for _MSVER<1900

hjmjohnson:
The ordering of the combined #if were causing an incorrect
overwriting of vcl_snprintf as std:;snprintf even in the
special case when _MSVER<1900.


hjmjohnson:
@thewtex This is the failure on the ITK dashboard.

PR #256, opened 2016-03-30
BUG: Generation of export header is unconditional

hjmjohnson:
The generation of the vul_export.h file was in a conditional part
of the cmake list based on platform type.  It is now pushed outside of
the platform definition section.


thewtex:
+1 Thanks for addressing this!

PR #257, opened 2016-03-30
Force constant api for external packages

hjmjohnson:


config/cmake/config/vxlGetCXXCompilerVersion.cxx
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+
+int main()
+{
+  #if __cplusplus >= VXL_CXX_TEST_VERSION
+    std::cout << __cplusplus << ">=" << VXL_CXX_TEST_VERSION << std::endl;
+    return 0;
```
DVigneault: Trivial point, but I know in ITK the `EXIT_FAILURE`/`EXIT_SUCCESS` macros are preferred--is the same true in VXL?

hjmjohnson: Correct.  This code is never run, and the return statement is not used/interpreted.  It does not indicate success or failure, so either value would be OK.

PR #258, opened 2016-04-03
ENH: Compose new matrix by row and column indices.

DVigneault:
This patch adds methods allowing a matrix to be composed from the rows/columns of a given matrix by supplying a vector of row/column indices.  This is conceptually analogous to what can be done in other languages/packages, e.g., in R:

    m <- matrix(data = c(1, 2, 3, 4, 5, 6), nrow = 2, ncol = 3)
    m.from.rows <- m[c(2,1,1,2),]
    m.from.cols <- m[,c(3,2,1,1,2,3)]

One specific use case might be in sorting eigenvectors based on eigenvalues.

    vnl_real_eigensystem eig(SomeMatrix);
    const auto eigenvectors = vnl_real(eig.V);
    const auto eigenvalues = vnl_real(eig.D.get_diagonal());
    
    TSort sort;
    vnl_vector<TReal>  sortedEigenvalues;
    vnl_vector<TIndex> sortedEigenvalueIndices;
    sort.vector_sort(eigenvalues, sortedEigenvalues, sortedEigenvalueIndices);
    const auto sortedEigenvectors = eigenvectors.get_columns(sortedEigenvalueIndices);

### DIFF
--- a/core/vnl/tests/test_matrix.cxx
+++ b/core/vnl/tests/test_matrix.cxx
@@ -95,6 +95,51 @@ void test_int()
 
   TEST("m2.put(1,1,3)", (m2.put(1,1,3),m2.get(1,1)), 3);
   TEST("m2.get(1,1)", m2.get(1,1), 3);
+
+  ///////////////////////////////
+  // Row and Column Operations //
+  ///////////////////////////////
+
+  {
+    int some_data[2*3] = { 1, 2, 3,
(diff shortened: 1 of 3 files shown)


DVigneault:
gcc only:

     The following tests FAILED:
        	853 - bmdl_pro_generate_mesh_process (OTHER_FAULT)

PR #261, opened 2016-04-05
Orbit fitting - incorporated changes from other orbit related branches and switched to new vxl

deuscovrigus:


hjmjohnson:
Failing test is unrelated to this patch set.

PR #271, opened 2016-04-15
Add option to compute rigid transform via vgl_compute_similarity_3d

decrispell:
This PR adds support for computing a rigid transform via the vgl_compute_similarity_3d class by fixing the scale at 1.   I thought that adding a separate "vgl_compute_rigid_3d" class (with the commonalities factored out into a third class) would be overkill but could be convinced otherwise.

### DIFF
--- a/core/vgl/algo/CMakeLists.txt
+++ b/core/vgl/algo/CMakeLists.txt
@@ -18,6 +18,7 @@ set( vgl_algo_sources
   vgl_norm_trans_2d.hxx                    vgl_norm_trans_2d.h
   vgl_norm_trans_3d.hxx                    vgl_norm_trans_3d.h
   vgl_compute_similarity_3d.hxx            vgl_compute_similarity_3d.h
+  vgl_compute_rigid_3d.hxx            vgl_compute_rigid_3d.h
                                            vgl_h_matrix_1d_compute.h
   vgl_h_matrix_1d_compute_linear.cxx       vgl_h_matrix_1d_compute_linear.h
   vgl_h_matrix_1d_compute_3point.cxx       vgl_h_matrix_1d_compute_3point.h
--- a/core/vgl/algo/Templates/compute_rigid_3d+double-.cxx
+++ b/core/vgl/algo/Templates/compute_rigid_3d+double-.cxx
@@ -1,0 +1,3 @@
+// Instantiation of vgl_compute_rigid_3d<double>
+#include <vgl/algo/vgl_compute_rigid_3d.hxx>
+VGL_COMPUTE_RIGID_3D_INSTANTIATE(double);
--- a/core/vgl/algo/tests/CMakeLists.txt
+++ b/core/vgl/algo/tests/CMakeLists.txt
@@ -4,6 +4,7 @@ add_executable( vgl_algo_test_all
   test_driver.cxx
 
   test_compute_similarity_3d.cxx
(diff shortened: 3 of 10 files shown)


mleotta:
Just my 2 cents, but I like the `vgl_compute_rigid_3d` approach better.  You don't really need a third class right?  Just make `vgl_compute_similarity_3d` factor out the scale and then call `vgl_compute_rigid_3d`.

decrispell:
I think the approach you describe would require the data to be preconditioned twice unnecessarily (once in vgl_compute_similarity_3d::estimate() and again in vgl_compute_rigid_3d::estimate()) since the points need to be centered before computing the scale in the similarity computation, and then need to be centered and scaled before computing the rotation in the rigid computation.   Probably not a big deal in practical terms, but it's a little ugly.

I was thinking that the preconditioning and rotation computations could be factored out and called by both vgl_compute_rigid_3d and vgl_compute_similarity_3d, but it was at that point that it started to feel like overkill.  Am I missing something?

mleotta:
No, you are right.  To avoid centering twice you would need to refactor things a bit.  Probably break up the estimate function into helper functions.  You still don't need another class. Just a little more modularity  of the functions.  One option would be to make `vgl_compute_rigid_3d` a parent class of `vgl_compute_similarity_3d` and move everything but the scale estimate into the rigid class.  You could break out the rotation/translation estimation function into protected helper functions that get called by both.  I don't think a separate `vgl_compute_rigid_3d` class is overkill, I think it's easier to understand and good design.  I do think a third class (e.g. `vgl_compute_rigid_3d_helper` or something like that) is overkill and ugly.

All this is just my opinion. There is nothing wrong with your current approach, it's just a little less obvious how you do rigid estimation to someone looking for that capability.

decrispell:
I agree that a vgl_compute_rigid_3d class will make the functionality easier to discover.  I don't particularly like the idea of using inheritance here (would estimate() then become a virtual function?  Do we really want to imply an "is a" relationship between rigid and similarity estimators?) , so the preconditioning will be done twice when estimating similarity transforms for now.

mleotta:
That's fine.

However, I'll point out that "is a" is not always the best way to think about C++ inheritance when it comes to mathematical concepts.  Sometimes that role is reversed because of what "is a" can mean mathematically.  In the space of geometric transforms "is a" often means "is a subset of".  A rigid transformation is a subset of similarity, which is a subset of affine, etc.  But to describe these concepts, the more superset class needs more parameters than the subset.  The C++ inheritance "is a" means more like "is an extension of".  So affine is an extension of similarity, which is an extension of rigid, etc. Also, `estimate()` would not need to be virtual.  You can have inheritance without any virtual functions.

All of that said, inheritance is not really necessary here.  Another options is making the `vgl_compute_rigid_3d` as a copy of `vgl_compute_similarity_3d` but without the scale stuff.  Then break the code into a few more protected member functions, then make `vgl_compute_similarity_3d` a friend class of `vgl_compute_rigid_3d` so that  `vgl_compute_similarity_3d` can contain a `vgl_compute_rigid_3d` and reused some of those helper functions.

decrispell:
Okay, all good points. In the interest of moving forward I'm going to merge the current solution.  It sounds we can move towards an ideal implementation as the need/motivation arises without changing the interface now that vgl_compute_rigid_3d is it's own class.

PR #272, opened 2016-04-20
Processes and bug fixes from voxel globe project

andyneff:
- Added/updated a number of processes, primarily focused on generating LE/CE statistics for a voxel world, generating point clouds, and geo registering cameras
- Fix `get_output_double` which affected ALL processes that returned double type, which were accidentally casting to float first.


andyneff:
The build only fails on tests that are known to randomly fail. Everything else looked good to me.

PR #274, opened 2016-04-27
Additional OpenCL reporting

drewgilliam:
Added additional OpenCL reporting to aid in multi-GPU debugging:
--PCI bus & slot numbers for NVIDIA devices
--Error number when OpenCL error is "unknown"


dstoup:
+1

PR #277, opened 2016-05-01
Add hidden visibility

hjmjohnson:


hjmjohnson:
1 failing test is not related to this patch set.

PR #279, opened 2016-05-01
Update to require c make289

hjmjohnson:


hjmjohnson:
@thewtex @decrispell @dstoup @TimCootes 

I'd like to update cmake to 2.8.9 to avoid bugs in 2.8.7 that would require a lot of work arounds.

Does anyone have an objection to this?

Hans

hjmjohnson:
The 1 test failure from clang is a test that often fails and is unrelated to this patch.

hjmjohnson:
@thewtex Thanks for the review Matt.  If I don't hear any objections, I'll merge soon.

Hans

dstoup:
Looks good to me!

PR #280, opened 2016-05-02
Fix missing space in flag append

judajake:
When appending fPIC to cxx and c flags there was a missing space, which cause errors/warnings when user provided extra flags


hjmjohnson:
👍  Thank you for the patch!

PR #283, opened 2016-05-03
BUG: VXL_CONSTEXPR requires initialization

hjmjohnson:
vxl/core/vnl/vnl_numeric_traits.h:182:41: error: constexpr static data member ‘maxval’ must have an initializer
   static VNL_EXPORT VXL_CONSTEXPR short maxval; // = 0x7fff;


hjmjohnson:
@blowekamp Your recent patch was causing build failures in C++11 builds.  This fixes the problem.

PR #288, opened 2016-05-04
STYLE: Remove old compiler work around for loops

hjmjohnson:
Remove unused macros for old style for loop scoping constructs.


hjmjohnson:
Failing test is not related.

PR #293, opened 2016-05-07
Better cxx11 granularity

hjmjohnson:


hjmjohnson:
@bradking @thewtex @blowekamp 

All,  I would appreciate a careful review of this set of patches.  I think that this approach would be a good one to replicate in ITK to provide better C++11 future support in a way that is granular and maintainable.

The approach uses cmake 3.5.2 and WriteCompilerDeetectionHeader ONE TIME to create files that are then statically included in the source tree (with very minimal modifications.. 4 lines of code added and a default file added for UNKNOWN compilers) so that the compiler detection is backwards compatible to cmake 2.8.9 minimum version.   

This paradigm should be very easy to maintain, and allows us to piggy back off the efforts of CMake developers to identify compiler feature detection.

Thank you for your considerations.

Hans

blowekamp:
Do any compiler exist which are 100% compliment to a standard? Sooner or later features will need to be broken down, and bugs detected.

hjmjohnson:
@blowekamp Thank you for reviewing.  The intent of this patch is to address the reality that compilers are not going to be 100% compliant and consistent across all environments.  This set of patches provides a mechanism for more fine grained control than the previous implementation.

bradking:
> The approach uses cmake 3.5.2 and WriteCompilerDeetectionHeader ONE TIME to create files that are then statically included in the source tree

Since this is a one-shot anyway, you could consider using cmake from our Git master branch.  In post-3.5 development we've added support for the Intel C++ compiler, at least on UNIX platforms.  I'm not sure what to do with that compiler on Windows because there doesn't seem to be enough information from the preprocessor to even know what `-Qstd=` option was given!

hjmjohnson:
@bradking THANK YOU Brad!  That is a great suggestion.  I will do that as a separate patch to help demonstrate how this can be accomplished in the future.

PR #294, opened 2016-05-09
Belief propagation algo for building voxel worlds

andyneff:
Added the belief propagation code and everything seems to be building/working.

Also included in this merge is a process to remove voxels with a low number of observations and a fix for Windows where ocl manager tries to unload too late, resulting in a pure virtual function call error (plaguing anyone in windows using the python adaptors to do anything ocl related).

@decrispell @ozgecanoz


hjmjohnson:
All tests pass, so merging.

PR #298, opened 2016-05-17
BUG: Address missing shared_ptr for VS9

blowekamp:
This patch addresses the following compilation error:

vxl\vcl\vcl_compiler.h(940): fatal error C1189: #error :  "Missing
definition for SHARED_PTR"

Older Visual Studio have tr1 include not in a tr1 directory but still
in the tr1 namespace, this patch add support for this configuration.


blowekamp:
This patch appeared to fix the problem building SimpleITK with ITK 4.10rc1. It was manually copied from that build.

hjmjohnson:
@blowekamp THANK YOU!

Issue #305, opened 2016-05-25
vnl std::complex instantiation problems with clang + libstdc++


rfabbri:
Hi,
I'm trying to build vxl on OSX using the default clang setup. It builds vnl OK, but since I need GNU extensions in my code, I activated the use of libstdc++ by

CMAKE_CXX_FLAGS = -Wno-gnu-static-float-init -stdlib=libstdc++

( Not sure if I should just switch to GCC entirely in this case. )

But then vnl spits out a link time error only for std::complex specializations of vnl structures:

```
[ 30%] Built target testlib
[ 30%] Linking CXX executable ../../../bin/vnl_test_all
Undefined symbols for architecture x86_64:
  "std::complex<double> dot_product<std::complex<double> >(vnl_vector<std::complex<double> > const&, vnl_vector<std::complex<double> > const&)", referenced from:
      test_vector() in test_complex.cxx.o
  "std::complex<double> inner_product<std::complex<double> >(vnl_vector<std::complex<double> > const&, vnl_vector<std::complex<double> > const&)", referenced from:
      test_vector() in test_complex.cxx.o
...
...
...
```

This did not occur about 6 months ago with the same CXX_FLAGS (when we'd use vcl_complex perhaps). I looked into the instantiation macros and can't find anything wrong. I did a clean rebuild, to no avail.

Thanks,

yanggehua:
Hi Ricardo,

I have been compiling VXL on MAC OS X as well, but not with the latest revisions.  Just a quick check:  did you try to use "-std=c++11"  in the compilation?  It may not help your situation at well. It is just the first idea I come up with looking at your message. 

Gary

hjmjohnson:
I just ran a clean build successfully:

 

mkdir vxl-libstdcxx

cd vxl-libstdcxx

cmake -DCMAKE_CXX_FLAGS:STRING="-Wno-gnu-static-float-init -stdlib=libstdc++" ../vxl

make

 

 

NOTE:  If you change the CMAKE_CXX_FLAGS after initial cmake configuration, you are effectively changing the compiler behavior AFTER all the introspection was done to determine what the compiler features are.  You MUST add these compiler-changing flags BEFORE the introspection is run (i.e. on the initial cmake command line).  

 

Flags like “-stdlib=libstdc++, -std=c++11” effectively change the compiler and all it’s behavior.

 

Hans

rfabbri:
Compilation works, but the problem persists with

BUILD_SHARED_LIBS = ON

[...]

PR #306, opened 2016-05-26
BUG: Missing VNL_EXPORTS for shared libs

hjmjohnson:
These missing VNL_EXPORTS were identified by
linker errors of unidentified symobols when
building a shared library.

cmake -DBUILD_SHARED_LIBS:BOOL=ON \
      -DCMAKE_CXX_FLAGS:STRING="-Wno-gnu-static-float-init -stdlib=libstdc++" \
      ../vxl

Special thanks to Ricardo Fabbri for identifying and being persistent narrowing
down the root scenario that exposed the problem.

resolves: #305


rfabbri:
excellent, thanks!

hjmjohnson:
Failure is unrelated.  It is one of the unstable tests: 661 - gevd_test_noise (OTHER_FAULT)

PR #307, opened 2016-05-26
support build of vpgl without geotiff

mleotta:
This commit makes it possible to build vpgl without GeoTIFF enabled.  The rest of core also builds without GeoTIFF, but there are numerous issues in BRL.


decrispell:
Merging this in, I will submit a separate PR that addresses the brl failures it triggers.

Issue #304, opened 2016-05-23
VXL does not build without GeoTIFF


mleotta:
VXL seems to mostly support GeoTIFF as an optional dependency as there are numerous checks for "HAS_GEOTIFF" or "GEOTIFF_FOUND".  However, VXL does not build without GeoTIFF.  

There are a few errors in core/vpgl/algo/vpgl_camera_convert.cxx that can be resolved by enclosing a couple of functions in `#if HAS_GEOTIFF`.  However, there is a lot more code in BRL that just assumes GeoTIFF is available.

decrispell:
I will look into this.  I'm surpised that this does not show up on the dashboard.  My dashboard system does have system geotiff libraries installed, but it would be surprising to me if all the other ones did as well.

dstoup:
When the system doesn't have geotiff, I believe VXL builds it on your
behalf, thus all works well. VXL_USE_GEOTIFF is enabled by default.

I have solved the BRL issue in our local repos by disabling the entire
package when geotiff is disabled. I tried a more surgical approach at one
point, but the vpgl stuff is throughout.


On Thu, May 26, 2016 at 9:37 AM, Daniel Crispell <notifications@github.com>
wrote:

> I will look into this. I'm surpised that this does not show up on the
> dashboard. My dashboard system does have system geotiff libraries
> installed, but it would be surprising to me if all the other ones did as
> well.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/vxl/vxl/issues/304#issuecomment-221872740>
>



-- 
David Stoup
R&D Engineer

Kitware, Inc.
28 Corporate Drive
Clifton Park, NY. 12065
518-881-4949 (W)
518-312-3946 (M)
518-371-4573 (F)

mleotta:
Yes @dstoup, you are right that GeoTIFF is in v3p and VXL will build it if not found.  The particular situation I found my self in what building on a machine that had system packages for both TIFF and GeoTIFF installed.  However, I needed to build VXL against a custom version of TIFF that I built (specifically the one provided by Fletch - https://github.com/kitware/fletch).  I did not have a custom version of GeoTIFF built against that custom TIFF, and VXL kept wanting to use my system GeoTIFF (which uses the wrong TIFF), so I tried to turn GeoTIFF off.  That's when I discovered that you can't really turn GeoTIFF off.

[...]

PR #308, opened 2016-05-27
Guard geotiff deps

decrispell:
Guard brl modules dependent on geotiff.  Addresses #304


DVigneault:
+1 -- Thanks for adding this!  Just built this PR, and it fixes a build failure on my system.

decrispell:
@mleotta This solutions is a bit heavy-handed (skip entire brl module that have any direct or indirect need for geotiff), but it should do the job.  Fixes #304

mleotta:
Thanks Dan

PR #312, opened 2016-06-01
Add sse abs diff

dstoup:
Add SSE optimizations to vil


core/vil/CMakeLists.txt
```diff
@@ -271,6 +271,11 @@ set( vil_sources ${vil_sources}
   # Plugin for image loading
   vil_image_resource_plugin.cxx         vil_image_resource_plugin.h
 )
+IF ( ${VXL_HAS_SSE2_HARDWARE_SUPPORT} )
```
hjmjohnson: Please use modern lower case cmake directives.

core/vil/CMakeLists.txt
```diff
@@ -271,6 +271,11 @@ set( vil_sources ${vil_sources}
...
+  SET( vil_sources ${vil_sources}
+    vil_math_sse.cxx vil_math_sse.h
+  )
+ENDIF ( ${VXL_HAS_SSE2_HARDWARE_SUPPORT} )
```
hjmjohnson: endif should not have elements in the  () in modern cmake.

PR #316, opened 2016-06-08
Add readme

decrispell:
Since we're using github, I think vxl should have a top level README.md.  I made a direct translation of the sourceforge VXL webpage as a starting point.  I also can take a pass at updating the info and removing dead links if there is interest in using this.


DVigneault:
@decrispell 👍   I think this will make the project a lot more inviting to people who stumble upon the GitHub.

PR #317, opened 2016-06-12
ENH: Encourage unified container interface.

DVigneault:
VNL containers are not related by inheritance; as such, maintaining
a common interface (where applicable) is not enforced.  Therefore,
container classes may diverge, causing methods to be absent which
the user would logically expect to have.  This patch is a very early
attempt at uncovering and implementing such missing methods through
test-driven-development.  In particular, the following four methods
have been added to `vnl_matrix_fixed`:

- `swap`
- `is_equal`
- `get_rows`
- `get_columns`

In addition to providing a (more) unified interface, this technique
may reduce the burden in writing unit tests, which currently are
written separately for each container, even when the behavior is
expected to be identical.  The `container_interface` unit test
which is added here provides runtime unit tests only for the
methods which were added (this will hopefully change in the future).

### DIFF
--- a/core/vnl/tests/test_matrix.cxx
+++ b/core/vnl/tests/test_matrix.cxx
@@ -160,18 +160,30 @@ void test_int()
        (m0.set_diagonal(vnl_vector<int>(2,2,m0values)),
         (m0.get(0,0)==7 && m0.get(1,1)==9 && m0.get(0,1)==2 && m0.get(1,0)==2)), true);
 
-  // Flip the matrix, otherwise it is the same whichever way it is flattened
-  m0.flipud();
-  // | 2 9 |
-  // | 7 2 |
-  vnl_vector<int> flat;
-  TEST("m0.flatten_row_major()",
-       (flat = m0.flatten_row_major(),
-       (flat.get(0)==2 && flat.get(1)==9 && flat.get(2)==7 && flat.get(3)==2)), true);
-  TEST("m0.flatten_column_major()",
-       (flat = m0.flatten_column_major(),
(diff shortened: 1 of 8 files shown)


DVigneault:
@hjmjohnson @thewtex I would appreciate your feedback on this.

hjmjohnson:
@DVigneault Very nice.  Thank you for the contribution.  

In the future, it would be great to propagate use of this common code to other containers, and then REMOVE existing redundant testing.

Thanks Hans.

core/vnl/tests/test_matrix.cxx
```diff
@@ -160,18 +160,30 @@ void test_int()
...
+                     8,  9, 10, 11,
+                    12, 13, 14, 15};
+
+    vnl_vector<int> flat(data, 16);
+
+    vnl_matrix<int> sq(data, 4, 4);
+    vnl_matrix<int> lg(data, 2, 8);
+    vnl_matrix<int> wd(data, 8, 2);
+
+    TEST("sq.flatten_row_major", flat.is_equal(sq.flatten_row_major(), 10e-6), true);
```
hjmjohnson: @DVigneault   Lines 179 - 185 are causing compilation failures on the intel compiler.

error: no suitable user-defined conversion from "vnl_vector_fixed<int, 16U>" to "const vnl_vector<int>" exists


Is this intended?

hjmjohnson: See #319

DVigneault:
@hjmjohnson Sorry for the delay, I've been out of lab for a few days.  `test_matrix_fixed.cxx:251` was certainly an error on my part--I'm surprised that g++ did not complain--but I think your solution is fine.  Thank you very much for taking action on this!

PR #318, opened 2016-06-16
COMP: Allow compilation for older compilers

hjmjohnson:
Older compilers may not be explicitly supported
in cmake for granular language support.
In these cases fall through to a default
behavior when using old compilers.


hjmjohnson:
@blowekamp Brad,  I backported a change to VXL, and will soon implement in ITK as well to allow for VS9 support.  Thank you for identifying this problem.

blowekamp:
Great!

PR #319, opened 2016-06-17
COMP: Intel compiler incompatibility

hjmjohnson:
The intel compiler identifited that:
error: no suitable user-defined conversion from "vnl_vector_fixed" to "const vnl_vector" exists

Do the conversion explicitly.  This also exercised more code and
required a change to avoid a linking problem with vnl_vector_fixed.


hjmjohnson:
@DVigneault I think this fixes the problem.

PR #329, opened 2016-06-24
Remove outdated compiler conditionals

hjmjohnson:
This is a first step in removing conditional code that was needed for old compilers.  In many cases, these conditional flags were never used in the VXL toolkit, it's contributions, or in ITK.


hjmjohnson:
VERIFIED that there are no compilation problems with MS visual studio 9.  (Warnings have been there for a long time) https://open.cdash.org/buildSummary.php?buildid=4444340

Also VERIFIED that other are no compilation problems with MS visual studio 15. (Warnings have been there for a long time)  https://open.cdash.org/buildSummary.php?buildid=4444381

bradking:
LGTM.

Instead of `"Compilers that do not support ... are not supported"` errors I think you could even just remove the checks altogether and hard-code the answers to true.  They probably don't even have to be something CMake is aware of and could just be hard-coded directly on the C++ side.  This could be done in a separate PR as a follow-up though.

hjmjohnson:
@bradking 

Thanks Brad.  I originally did remove that code too, but I thought that having the introspection out in the wild a little while for a little bit might identify compilers that are not normally tested and give the end-user a really good message about why VXL is not working on their platform.

I will remove that remaining code in a few months from now.  By removing those extreme sanity checks, it should improve initial cmake configuration times.

yanggehua:
@hjmjohnson looks good to me.

PR #334, opened 2016-07-01
Fix building shared libraries on Windows

msmolens:
When building shared libraries, Windows must not have the export specification for templated class declarations, while GCC must have the export specification. To enforce this, add a local copy of GenerateExportHeader that provides the &lt;module-name&gt;_TEMPLATE_EXPORT macro and use the the macro where necessary.

This is intended to fix failures on ITK dashboard, e.g. https://open.cdash.org/viewBuildError.php?buildid=4437123. Building vxl standalone with shared libraries on Windows may still be broken.

_Edit:_
Note that the VNLInstantiation module in ITK also requires changes to be compatible. See proposed fix: https://github.com/msmolens/ITK/commit/b4ae18c8e0cb5e7cafa2a28dc1c563c341e31644


msmolens:
cc: @thewtex @blowekamp

thewtex:
+1

@hjmjohnson @bradking @dstoup Please review and merge so we can test the result on the ITK dashboard.

dstoup:
I also mentioned this request to Bill since he contributed the original
win32 shared support. Would be good to get his opinion. I will test the
branch on Tuesday.

hjmjohnson:
👍  Thank you for the contribution.

jcfr:
:+1:  Very 

@hjmjohnson @decrispell @yidongVSI  Since CI is green let us know what is missing to finalize the integration.

jcfr:
Thanks @hjmjohnson

PR #336, opened 2016-07-06
COMP: Enable hidden visibility property with vxl static libraries

blowekamp:
Set CMP0063 to NEW to enable the CMake visibility properties to be
honored when building static libraries.


thewtex:
:+1:

hjmjohnson:
Thanks Brad!

PR #342, opened 2016-07-12
Txx to hxx

yanggehua:
1) Rename *.txx files to .hxx in contrib/rpl
2) Replace ".txx" occurrence in core to ".hxx", which seems being left-out in the previous change


hjmjohnson:
👍  This has been on my todo list for a long time!  Thanks.

@thewtex

yanggehua:
@hjmjohnson  I wasn't sure where is the boundary for this change.  Am I correct that all vcl and core shall have been using .hxx instead of .txx?  As for the contrib, it is up to each group to make the change?

thewtex:
Thanks for taking the time to make this change. Great to see vxl get some TLC for this update.

Issue #343, opened 2016-07-13
Crash in FFMpeg when calling vidl_ffmpeg_istream:: seek_frame(unsigned int frame)


yanggehua:
I was trying to seek to frame 420 on a avi file that is encoded with RGBP codec (Palettized RGB). Essentially every frame is a key frame with this codec.  But the call to vidl_ffmpeg_istream:: seek_frame(unsigned int frame) crashed in the middle of call to avcodec_flush_buffers().

If I commented out the call to avcodec_flush_buffers() (line 620 of vidl_ffmpeg_istream_v56.hxx).  It actually worked!  But when I look at the documentation, it clearly stated "Reset the internal decoder state / flush internal buffers. Should be called e.g. when seeking or when switching to a different stream." (http://ffmpeg.org/doxygen/trunk/group__lavc__misc.html#gaf60b0e076f822abcb2700eb601d352a6).  

I wonder if anyone has experience on this FFMpeg issue.  I guess it is possible that it doesn't need to be "flushed" as no frame was read and the first action was seek_frame.  

FFMPeg version detail is as follows.  The vidl ffmpeg file being used is vidl_ffmpeg_istream_v56.hxx.  

FFmpeg version: 20160105-git-68eb208
  libavutil      54. 30.100 / 54. 30.100
  libavcodec     56. 57.100 / 56. 57.100
  libavformat    56. 40.101 / 56. 40.101
  libavdevice    56.  4.100 / 56.  4.100
  libavfilter     5. 32.100 /  5. 32.100
  libswscale      3.  1.101 /  3.  1.101
  libswresample   1.  2.101 /  1.  2.101
  libpostproc    53.  3.100 / 53.  3.100

yanggehua:
@mleotta I wonder if you have any knowledge about this "flush" usage. :)

mleotta:
@yanggehua I do not have any other insight here.  From my experience the "seek" functionality in FFMPEG has always been unreliable, but I've never had a crash as a result.  Usually the problem is that seek doesn't give you frame-accurate seeking because it is based on time stamps rather than frame numbers.  Time stamps are usually sufficient if the end application is video playback, but not so much if you need to find an exact frame for a vision application.  Our usual workaround is to advance through the video frame-by-frame and count frames.  This is not too expensive if you advance each frame but do not decode it.

yanggehua:
[...]

PR #344, opened 2016-07-13
fixed the crash in seek_frame().

yanggehua:
Essentially, codec must be checked for validity before calling avcodec_flush_buffers() function. Or empty internal pointer will lead to hard crash.


yanggehua:
discussed in Issue #343

PR #346, opened 2016-07-21
Fixed the install path of vnl_config.h

mleotta:
Previously it was installed to vxl/vnl/vnl_config.h while it should
have been vxl/core/vnl/vnl_config.h

### DIFF
--- a/core/vnl/CMakeLists.txt
+++ b/core/vnl/CMakeLists.txt
@@ -74,7 +74,9 @@ else()
   set(VNL_CONFIG_ENABLE_SSE2_ROUNDING 0)
 endif()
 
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@thewtex Matt there have been recent changes merged in VXL.   Will these conflict with your changes for backwards compatibility in ITK?

mleotta:
These two changes I made seemed like pretty obvious errors, which is why I merged them so quickly.  I can't imagine that fixing them would break ITK.  In fact, I found them because I was getting errors building a 3rd party package against installed VXL (it built fine against the VXL build tree, but not against installed VXL).  These two vnl configured files were missing the `core` in their install path, making their location inconsistent with all other installed files, including similar configured header files in other core libraries.

thewtex:
ITK historically has different header installation locations which need to be maintained for backwards compatibility. This changes will cause build failures when someone tries to build against ITK. We want to minimized any forked code so ITK and VXL and push and pull patches without conflicts. Here is a patch that adds ITK-compatible locations:
  
#348

PR #347, opened 2016-07-22
Fixed install path for vnl_algo_export.h

mleotta:
Previously it was installed into vxl/vnl/algo/vnl_algo_export.h
it should be vxl/core/vnl/algo/vnl_algo_export.h


hjmjohnson:
@thewtex Here is another change.

PR #368, opened 2017-01-31
Fix linking issues with shared builds on VS 2015

billhoffman:
This solves some undefined symbols errors seen when linking shared on VS 2015.


billhoffman:
@dstoup take a look.

PR #373, opened 2017-02-03
add new affine coordinate functions

mundyj:
Added affine coordinate functions to vgl


hjmjohnson:
@mundyj Would you please rebase on the latest master branch?  We have addressed the long-running test failures, and it would be nice to see passing dashboard before reviewing this patch set.

decrispell:
@hjmjohnson Done.

PR #377, opened 2017-02-08
Let users decide if they should include DCMTK.

dstoup:
VXL doesn't build against all versions of DCMTK which makes the 'always use' approach VXL takes, problematic.


hjmjohnson:
@dstoup 👍

PR #378, opened 2017-02-24
Old fix for third order missing

rfabbri:
We're heavily developing the edge detectors,
and noticed this bug which was causing all edge orientations to be off.

This is tested and have been in agreeement with Yuliang Guo from LEMS/Brown,
the main researcher improving edge detection. I'm the main vxl maintainer of this edge feature system and a main user of it.


hjmjohnson:
👍

PR #379, opened 2017-03-01
VXL: dbsol to bsold_circ_arc in vsol curve

rfabbri:
This code isn't great, as it cites external classes, but it needs this fix since
dbsol_circ_arc_2d doesn't exist anymore in any recent active repository (this was an old LEMS/Brown class). Only bsold_circ_arc_2d does, and this is from the VXD library http://github.com/rfabbri/vxd.

(cherry picked from internal commit lemsvpe@daa998a4616ec57902357898b0ba30614016538e)

### DIFF
--- a/contrib/gel/vsol/vsol_curve_2d.h
+++ b/contrib/gel/vsol/vsol_curve_2d.h
@@ -28,7 +28,7 @@ class vsol_line_2d;
 class vsol_conic_2d;
 class vsol_polyline_2d;
 class vsol_digital_curve_2d;
-class dbsol_circ_arc_2d;
+class bsold_circ_arc_2d;
 
 class vsol_curve_2d : public vsol_spatial_object_2d
 {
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@rfabbri Can you comment on this patch set a bit more?  Is this a work in progress?  Who will be affected by these changes.

After so many months, perhaps this is not a good change.

rfabbri:
This is still needed for for anyone using a codebase from Brown.
Since this has been open-sourced into VXD http://github.com/rfabbri/vxd, the new class is now bsold_circ_arc_2d, so this must be done.

People that will be affected is people using an old Brown codebase; however, these people cannot compile against the latest VXL for a long time. The required changes to use the latest VXL are all in VXD anyways.

This is not work in progress; eventhough I have not announced this library http://github.com/rfabbri/vxd, it is already functional and people have been using it.

PR #382, opened 2017-03-01
Sdet minor

rfabbri:
minor fixes to include paths and remove empty file.


rfabbri:
The build failure is nothing related to these changes. It is just a random unrelated failure in gel/gevd/tests that showed up for this particular build under Clang. Therefore, I suggest this commit can be safely merged despite this failing build.

PR #389, opened 2017-05-28
Set initial value for best_x.

vicory:
This will prevent errors caused by returning an empty vector when starting at a minimum.


hjmjohnson:
👍

PR #392, opened 2017-06-22
Add runtime-loaded formats to the front so they can replace built-in formats.

dstoup:
We need to add custom, runtime generated image formats to the head of the vil_file_format_storage list. The return of a matching format happens first come, first served, which makes it impossible to use custom formats which extend, built-in types.


dstoup:
The travis failure is not related to this change. It is an unrelated failure in gel/gevd/test that shows up sometimes under Clang. This commit can be merged safely.

PR #399, opened 2017-07-04
openmvs

dxqxiaoqiang:
不错


hjmjohnson:
@dxqxiaoqiang This change is not justified.  It appears that a single new line was added.

PR #404, opened 2017-09-01
Remove outdated policies

hjmjohnson:


hjmjohnson:
The failing test seems unrelated to this patch set:
The following tests FAILED:
	 37 - tiff_raw_decode (Failed)
Errors while running CTest

hjmjohnson:
@thewtex Matt,  I'm making a push so that we can update ITK to remove warnings before the next release.   This will help ITK build without cmake warnings.

PR #405, opened 2017-09-02
BUG: Bad test for optionally built tiff_raw_decode

hjmjohnson:
From the tiff 4.0.8 version of the test:

 JPEG decoding is inherently inexact, so we can't test for exact
 pixel values.  (Well, if we knew exactly which libjpeg version
 we were using, and with what settings, we could expect specific
 values ... but it's not worth the trouble to keep track of.)
 Hence, use ranges of expected values.  The ranges may need to be
 widened over time as more versions of libjpeg appear.

The ranges specified in the new test is EXACTLY the difference
between the previous failing test and previously required
exact values.

### DIFF
--- a/v3p/tiff/test/CMakeLists.txt
+++ b/v3p/tiff/test/CMakeLists.txt
@@ -1,5 +1,6 @@
 # v3p/tiff/test/CMakeLists.txt
 
+
 # Include the generated headers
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@thewtex, @dstoup ,

 Could you review this patch please?

PR #407, opened 2017-09-02
Fix forward declarations

hjmjohnson:


hjmjohnson:
@thewtex Matt,  I'm making a push so that we can update ITK to remove warnings before the next release.

PR #414, opened 2017-11-25
COMP: Fix linking inconsistenies

hjmjohnson:
Linking warnings are pervasive in ITK and other
packages due to inconsistent visibilty settings.

ld: warning: direct access in function 'foo()' from file 'foo.o' to global weak symbol 'vtable for vnl_svd<float>' from file '../../../../lib/libvnl_algo-4.13.a(vnl_svd+float-.cxx.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.

This was an oversight in a previous refactoring.


hjmjohnson:
Failing test seems to have nothing to do with this change:

The following tests FAILED:
	375 - vpgl_algo_test_bundle_adjust (Failed)
Errors while running CTest

dstoup:
LGTM

PR #416, opened 2017-12-01
With hidden visibility propagated we need to export symbols

dstoup:
With visibility set to hidden applications can't link vnl_algo. This branch exports the classes that were failing for my applications. It's likely that a more comprehensive branch should follow which exports more of this libraries symbols.

### DIFF
--- a/core/vnl/algo/vnl_amoeba.h
+++ b/core/vnl/algo/vnl_amoeba.h
@@ -23,6 +23,8 @@
 #include <vnl/vnl_vector.h>
 #include <vnl/algo/vnl_algo_export.h>
 
+#include <vnl/algo/vnl_algo_export.h>
+
 class vnl_cost_function;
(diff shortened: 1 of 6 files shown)


hjmjohnson:
@dstoup THANKS!  I'll be happy to help make further changes if you want to give me a heads up on how to induce these problems.

dstoup:
It's possible that we can stop with this branch. I guess it depends on usage. I addressed all of our link failures and I suspect ITK isn't having this issue. For me, the issue is exposed by building VXL with hidden visibility then just trying to create/use objects in the vnl_algo classes. I guess it depends on which philosophy we subscribe to. On one hand I can see where all vnl_algo should be exported since it's probably meant to be used publicly. On the other, I can see keeping changes minimized to problems. There are a handful of vnl_algo classes which are not decorated. I just don't have a good sense if the should be or not.

PR #418, opened 2017-12-03
COMP: Adding exports for vnl_algo

hjmjohnson:
Add missing set of exports needed when compilng
with shared libraries for vnl_algo.


dstoup:
Excellent, thank you. Not sure why i didnt try that, figured it needed the template_export. Merging, then will rerun dashboards that I control.

PR #419, opened 2017-12-05
COMP: Fix VXL_NOEXCEPT macro.

phcerdan:
Remove VXL_NOEXCEPT_EXPR, invalid in c++03.
Use empty `throw()` if noexcept specifier is not available.

Originated from ITK, see gerrit: http://review.source.kitware.com/22861

### DIFF
--- a/vcl/vcl_compiler_detection.h
+++ b/vcl/vcl_compiler_detection.h
@@ -249,10 +249,8 @@
 
 #  if VXL_COMPILER_CXX_NOEXCEPT
 #    define VXL_NOEXCEPT noexcept
-#    define VXL_NOEXCEPT_EXPR(X) noexcept(X)
 #  else
-#    define VXL_NOEXCEPT
-#    define VXL_NOEXCEPT_EXPR(X)
+#    define VXL_NOEXCEPT throw()
 #  endif
 
 


phcerdan:
Hi @hjmjohnson, `NOEXCEPT_EXPR` macros are valid in c++11 but I think they generate misleads when expanded in c++03. Let me show an example to illustrate it.

Also, note that the macros with `_EXPR` are not currently used in VXL or ITK.
And the behavior that I am proposing for ITK_NOEXCEPT in the patch is provided by the extra macro ITK_NOEXCEPT_OR_THROW, which name I find extra unfortunate for the user/developer and is the reason that motivated this PR in ITK.

`./Modules/Core/Common/include/itk_compiler_detection.h`.  the same change has been proposed, I shouldn't have opened this in parallel to that, but those headers are generated automatically from the same file in VXL and ITK. Please see: [gerrit](http://review.source.kitware.com/22861) and [discourse](https://discourse.itk.org/t/warning-dynamic-exception-specifications-are-deprecated-in-c-11/470/37) discussion.

The example: https://godbolt.org/g/M8Y6Lw

```cpp
#include <iostream>

#if __cplusplus >= 201103L
#define ITK_NOEXCEPT noexcept
#else
// ITK_NOEXCEPT in c++03 expands to throw() with this patch
#define ITK_NOEXCEPT throw()
#endif

/*
 * noexcept is already equivalent to noexcept(true).
 * There is no use of this macro if c++11 is not a requirement.
 * - macro won't be used in c++03:
 *  - It would lead to compiler errors if used as the valid noexcept operator in c++11
*/
#if __cplusplus >= 201103L
#define ITK_NOEXCEPT_EXPR(X) noexcept(X)
#else
#define ITK_NOEXCEPT_EXPR(X)
#endif

void no_throw() ITK_NOEXCEPT;
void may_throw();
// Both will be expanded to nothing in c++03 (both functions throwing), independently of the boolean.
void may_throw_with_expr_qualifier() ITK_NOEXCEPT_EXPR(false);
void no_throw_with_expr_qualifier() ITK_NOEXCEPT_EXPR(true);

int main() {
   std::cout << "Check differences of ITK_NOEXCEPT_EXPR between c++11 and c++03/c++98 \n";
   // This is valid c++11, but unintended use of the macro as an operator:
   std::cout << std::boolalpha << "may_throw() is noexcept? " << ITK_NOEXCEPT_EXPR(may_throw()) << "\n";
   std::cout << std::boolalpha << "no_throw() is noexcept? " << ITK_NOEXCEPT_EXPR(no_throw()) << "\n";
   std::cout << std::boolalpha << "may_throw_with_expr_qualifier() is noexcept? " << ITK_NOEXCEPT_EXPR(may_throw_with_expr_qualifier()) << "\n";
   std::cout << std::boolalpha << "no_throw_with_expr_qualifier() is noexcept? " << ITK_NOEXCEPT_EXPR(no_throw_with_expr_qualifier()) << "\n";
   
}
/* c++11 out:
Check differences of ITK_NOEXCEPT_EXPR between c++11 and c++03/c++98
may_throw() is noexcept? false
no_throw() is noexcept? true
may_throw_with_expr_qualifier() is noexcept? false
no_throw_with_expr_qualifier() is noexcept? true
 */

/* c++03 doesn't compile because use of macro as operator
 * throw() doesn't provide a static check, but all the functions
 * will throw with the current value of macros.
 */
```

I understand changing macros is not optimal though. And maybe all this hassle for logical correctness in c++03 can be omitted if c++11 is a requirement. But I hope you understand the motivation behind the patch.

PR #420, opened 2017-12-06
ENH: VNL_ALGO_EXPORT for advanced fft functions

hjmjohnson:
When compiling ITK, low level fft functions were identified
that need to be externally visible.

The vnl_fft_gpfa must be visible in shared library builds.

### DIFF
--- a/core/vnl/algo/vnl_fft.h
+++ b/core/vnl/algo/vnl_fft.h
@@ -9,11 +9,12 @@
 // \author fsm
 
 #include <vcl_compiler.h>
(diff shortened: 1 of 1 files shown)


dstoup:
Good find, thanks!

hjmjohnson:
@dstoup I think this is the last item needed for ITK to build against this version of VXL.

dstoup:
Excellent and I think our stuff is all building now too. I would say merge anytime. I don't know if the CI tests ever pass anymore because of timeout, but ...

PR #421, opened 2017-12-11
BUG: Allow compilation of VNL with MSVC<1600

fbudin69500:
Compiler detection and feature support is performed with files generated with
the CMake macro `WriteCompilerDetectionHeader`. This macro generate a specific
header file for several supported compilers, including MSVC. However, for
MSVC, it only supports feature detection if MSVC >= 1600. If MSVC < 1600,
an error message is thrown instead of defaulting to default unknown compiler
which uses `__cplusplus` variable to configure supported features. In the case
of MSVC < 1600, it will default to a configuration with no C++11 support.

### DIFF
--- a/vcl/vcl_compiler_detection.h
+++ b/vcl/vcl_compiler_detection.h
@@ -128,7 +128,7 @@
 # undef VXL_COMPILER_IS_GNU
 # define VXL_COMPILER_IS_GNU 1
 
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && _MSC_VER >= 1600
 # undef VXL_COMPILER_IS_MSVC
 # define VXL_COMPILER_IS_MSVC 1
 


fbudin69500:
@hjmjohnson This should allow to build VXL (and by extension ITK) with VS2008. It is necessary to support VS2008 for ITK to support Python 2.7 on Windows. This fix has been merged in ITK to verify that it solves the [compilation issue](https://open.cdash.org/viewBuildError.php?buildid=5176320) but VXL can be synchronized with ITK once this patch is merged here.

hjmjohnson:
@fbudin69500 Thank you so much for tracking this down.

thewtex:
:+1:

fbudin69500:
Looks like it solved the issue on the ITK dashboard. The 50 build errors related to this have disappeared: https://open.cdash.org/index.php?project=Insight

thewtex:
:fireworks: Thanks again, @fbudin69500

PR #423, opened 2017-12-12
COMP: Do not use export macro in vnl_amoeba default_verbose

hjmjohnson:
This avoids the build error

vnl\src\vxl\core\vnl\algo\vnl_amoeba.h(106)
: error C2487: 'default_verbose' : member of dll interface class may not be
declared with dll interface

when BUILD_SHARED_LIBS=ON and CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS:BOOL=ON are
set following the addition of the export macro to the vnl_amoeba class in vnl.


hjmjohnson:
@thewtex Thanks for fixing this in ITK.

PR #425, opened 2017-12-18
STYLE: Use VTK_NULLPTR to utilize c++11 features

hjmjohnson:
Where available, use c++11 features to identify
potential nullptr usage errors.


dstoup:
Looks reasonable to me.

PR #426, opened 2018-01-06
COMP: Use C++ headers over C headers

hjmjohnson:
Some headers from C library were deprecated in C++ and are no longer
welcome in C++ codebases. Some have no effect in C++. For more details
refer to the C++ 14 Standard [depr.c.headers] section.

This patch replaces C standard library headers with their C++
alternatives and removes redundant ones.

### DIFF
--- a/contrib/brl/b3p/expatpp/expatpp.cpp
+++ b/contrib/brl/b3p/expatpp/expatpp.cpp
@@ -8,7 +8,7 @@
   #include <cstring> // for std::strlen() & strcmp()
   #include <string>
   using namespace std;
-  #include <assert.h>
+  #include <cassert>
 #endif
 #include "expatpp.h"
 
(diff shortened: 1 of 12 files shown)


dstoup:
LGTM

dstoup:
Since this branch landed, my Mac dashboards are having issues with duplicate definitions of strstr, etc from string.h and abs, etc from stdlib.h

The errors are coming from core/vpl/os_unix/vpl.cxx. I removed cstring and cstdlib from the extern C block which fixes the issue for me. That change feels right to me since they are C++ headers, not C, but wanted to ping you here to make sure that makes sense to you as well. Thoughts?

dstoup:
I went ahead and created the MR anyway ... 
https://github.com/vxl/vxl/pull/428

hjmjohnson:
Yes.   That is correct.  Sorry,  I must have missed a few of these.

dstoup:
No problem vpl isn't often turned on. I will go ahead and merge that request.

PR #427, opened 2018-01-06
STYLE: macro definition with () protected arguments

hjmjohnson:
https://clang.llvm.org/extra/clang-tidy/checks/misc-macro-parentheses.html

Finds macros that can have unexpected behaviour due to missing
parentheses.  Macros are expanded by the preprocessor as-is. As a
result, there can be unexpected behaviour; operators may be evaluated in
unexpected order and unary operators may become binary operators, etc.
When the replacement list has an expression, it is recommended to
surround it with parentheses. This ensures that the macro result is
evaluated completely before it is used.

It is also recommended to surround macro arguments in the replacement
list with parentheses. This ensures that the argument value is
calculated properly.

### DIFF
--- a/core/vbl/vbl_array_3d.h
+++ b/core/vbl/vbl_array_3d.h
@@ -29,8 +29,8 @@
 # define RANGECHECK(i,j,k) ((void)0)
 #else
 # include <vcl_cassert.h>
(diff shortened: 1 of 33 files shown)


hjmjohnson:
1 failing test only on clang seems completely unrelated to this MR.
The following tests FAILED:
	666 - gevd_test_noise (SEGFAULT)

PR #437, opened 2018-01-26
Set visibility on mac correctly

hjmjohnson:


hjmjohnson:
@thewtex @bradking @dstoup @DVigneault 

https://public.kitware.com/pipermail/insight-developers/2016-April/024826.html

I was always looking in the wrong place.    As a last ditch effort, I decided to try to update VXL to require cmake 3.9.5, and when cleaning out all the old outdated conditional code I stumbled across the answer.

PR #439, opened 2018-01-27
Update minimum CMake version

hjmjohnson:
Update to modern CMake version to simplify conditional build options.


andyneff:
CentOS/RHEL 7's native cmake is 2.8.12.
Amazon Linux too, is cmake 2.8.12

With the `epel-release` repo, CentOS and RHEL 7 can get up to 3.6.3, but not 3.9.5
For some reason, amazonlinux with `epel-release` uses CentOS 6's epel-release, which has no cmake3 available.

This is worth considering as RHEL, CentOS, and amazon are important use cases of vxl. I believe they can still download linux binaries and use them manually, but this complicates things for many users.

hjmjohnson:
@andyneff  Thank you for the response.

There has recently been a large debate on the ITK site:
https://discourse.itk.org/t/minimum-cmake-version-update/585

===============
I recognize that complications for many users arise from moving to 3.9.5
I also recognize that many complications arise for users staying with older versions of cmake as well.

Given that ITK (3.9.5), Slicer(3.9.5), BRAINSTools(3.9.5), and VTK(3.3.0 or 3.7.0) are moving to more modern versions of cmake, and that VXL's build environment can be greatly simplified by using newer version of cmake, I propose we move into the future at this point.

=================
My primary motivation for this move forward is that VXL is primarily being maintained and improved by developers on the aforementioned projects.  The effort needed to incorporate VXL in those other projects is becoming burdensome with different build requirements and backwards compatibility.

==================
Of particular interest is support of building with C++11, which is much easier to configure and use with cmake version 3.8.2 or greater.  We could put in conditional code to require 3.9.5 only when building with C++11.

==================
Could you support cmake version 3.3.0 for VXL?  At least then much of the conditional compilation code can go away.

if(CMAKE_CXX_STANDARD EQUALS "11" OR "14" OR "17" OR "20" )
   cmake_min_required(  3.9.5 )
else
   cmake_min_required( 3.3.0 )
endif

andyneff:
@hjmjohnson That sounds like a much better compromise.

hjmjohnson:
@andyneff Thank you for the review.  Suggested updates committed.  I would appreciate your review.

PR #445, opened 2018-02-13
Fix a bug introduced in reading vgl_point_2d

dstoup:
The stream started caring about < > but it wasn't testing the end for the end of stream before peeking again which was returning an invalid stream. I added treatment of < > the same as ( ).

I also reformatted the function so it was readable. Before, it was incredibly difficult to follow the flow.

### DIFF
--- a/core/vgl/vgl_point_2d.hxx
+++ b/core/vgl/vgl_point_2d.hxx
@@ -61,27 +61,57 @@ std::istream& vgl_point_2d<Type>::read(std::istream& is)
 {
   if (! is.good()) return is; // (TODO: should throw an exception)
   bool paren = false;
+  bool angle = false;
   Type tx, ty;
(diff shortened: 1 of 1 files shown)


dstoup:
Thanks Hans! I think that covers all of the issues we're currently having, though I can't build brl anymore. I might poke at those errors so the dashboards are green again (slowly) but I am much less concerned about that library.

hjmjohnson:
@dstoup Is there someone who is interested in BRL?  It's a bit of a burden that this 'contrib' package makes keeping the core healthy so cumbersome.

dstoup:
I don't know, we used to care about 5 files, but shed our dependence on them. I think just Joe and the Brown crew. I suppose I should disable it on most of my dashboards just to have some expected green.

PR #446, opened 2018-02-13
Provide a test for the bug reading vgl_point_2d

dstoup:
I verified the test breaks with the commit prior to my fix and that this test resolves the issue.


dstoup:
FYI, I verified the test breaks with the commit prior to my fix and that this test resolves the issue.

PR #447, opened 2018-02-21
add large penalty for negative focal length in camera optimization

decrispell:
This fixes a bug in which vpgl_optimize_camera::opt_orient_pos_f() would occasionally return a camera with a negative focal length.

There was already a similar check within the cost function used by vpgl_optimize_camera::opt_orient_pos_cal, but not in the position and focal length only variant.


decrispell:
Travis errors seem related to #446

hjmjohnson:
@dstoup,
  @decrispell  is reporting that a#446 appears to be causing a dashboard failure.  Does the test need updating?

hjmjohnson:
Rebased and verified build locally.

PR #456, opened 2018-02-26
ENH: Disabling VNL_CONFIG_ENABLE_SSE2 as a valid option

hjmjohnson:
Turning this flag on causes many instability problems on many platforms.

Test begin failing and segmentation faults occur in the core of ITK.


hjmjohnson:
@thewtex Matt, you were the last person to add anything substantive to the SSE2 code.  

All my testing seems to indicate that there are severe bugs or incompatibilities with this code.

Both Mac and Linux are failing with segmentation faults and sporadic failures.  Do you have any insights about what should happen to this code?  I'm tempted to remove all the broken code from the library.

Hans

thewtex:
@hjmjohnson Could you please create an issue and assign it to me?

PR #481, opened 2018-05-15
COMP: Add gcc8 support to vcl_compiler.

phcerdan:
It just adds a if/else branch to the existing logic.

Triggered here: https://discourse.itk.org/t/building-and-compiling-with-gcc-8-1/934/4


phcerdan:
Tested locally with gcc8 and it works.

PR #485, opened 2018-05-24
mbl_cloneables_factory: split into def and implementaiton

tjcorona:
The use of the MBL_CLONEABLES_FACTORY_INSTANTIATE() macro after the
implementation of mbl_cloneables_factory was causing the compilation error
"error: explicit specialization of 'objects' after instantiation". Breaking
the definition and implementation into two files fixes this issue.


tjcorona:
@dstoup Would you mind reviewing this merge request? The fix is a small one, but unfortunately it propagates to several files in the project.

PR #486, opened 2018-05-24
rgtl: include octree_cell_geometry.h/hxx when installing.

tjcorona:


tjcorona:
@dstoup Would you mind reviewing this merge request? For some reason these two files were omitted when installing rgtl.

PR #506, opened 2018-10-16
Install the vnl_fortran_copy_fixed headers. There are external files …

dstoup:
…that need them.


dstoup:
@thewtex this fixes the issue I was having while building ITK with an external VXL.

thewtex:
Thanks @dstoup . LGTM

PR #509, opened 2018-10-21
COMP: BRL is not compatible with shared library builds

hjmjohnson:
When building BRL, several circular dependencies and interlibrary
dependencies are not compatible with building with shared
libraries.


decrispell:
I will take a look at this to see if there is an easy way to resolve the circular dependency.  If not, this looks like a fine solution.

hjmjohnson:
@decrispell Thanks for looking at fixing the circular dependency.  I don't think it will take much effort for someone who is familiar with the code.

I want to focus my efforts elsewhere (i.e. the VNL core).

Hans

hjmjohnson:
Superceeded by https://github.com/vxl/vxl/pull/518

PR #510, opened 2018-10-21
Cxx11 code updates

hjmjohnson:
In this first proposed release, it highlights a proposal to VXL's adoption of and requirement for the C++11 standard. Prior to this proposal, a subset of C++11 functionality was utilized, when available, through backports and macros. VXL's 2.0.0 proposal deprecates or removes these macros (e.g. VXL_NULLPTR, VXL_DELETED_FUNCTION, VXL_CONSTEXPR, and VXL_OVERRIDE) and directly uses C++11 keywords such as delete, constexpr, nullptr, override, noexcept.


hjmjohnson:
I would appreciate comments concerning making a VXL 2.0.0 release that ONLY supports C++11 and great compilers.

hjmjohnson:
@dzenanz @thewtex 

I have been updating VXL to bring it in line with modern programming practices.  There has been no response to requests for comment on this move.

I would appreciate if one of you could review the progress made and comment on any pitfalls/concerns you have with these updates.  

PS:  The ITK team may want to consider forking from VXL to allow us to update VNL core to continue to support ITK without the baggage of keeping backward compatibility pre C++11 compilers.

TimCootes:
I think modernizing the code like this is a great idea - thanks for all your effort on this.
I've not had a chance to check it out to find out how it interacts with out local code yet - I'll try to do that soon.

hjmjohnson:
@TimCootes @mleotta Thanks for your feedback.  These changes will provide significantly improved alignment for supporting the ITK toolkit, so I'd really like to get this into the master branch.  

Even though the changes touch many files, the refactorings are mostly safe for "forward compatibility" (i.e. compiling old code with new compilers).  

There is a laundry list of items to address that are more susceptible to breaking compatibility that will involve changes to the cmake files.  I've purposefully avoided those changes for now.

My plan is to merge this into the master branch this weekend, unless I there is a request to hold off longer.

contrib/gel/gevd/gevd_noise.h
```diff
@@ -60,7 +60,7 @@ class gevd_noise
   static bool RealZeroCrossing(const float* dhist, const int nbin,
                                float& index);
  private:
-  gevd_noise() {} // no default constructor!
+  gevd_noise() = default; // no default constructor!
```
dzenanz: This might be confusing to someone reading this later on 😄

core/vcsl/vcsl_geocentric.h
```diff
@@ -25,14 +25,14 @@
 class vcsl_geocentric
   : public vcsl_geographic
 {
+ public:
   //***************************************************************************
   // Constructors/Destructor
   //***************************************************************************
 
   // Default constructor.
   vcsl_geocentric();
```
dzenanz: This constructor is now public. Was this intended? Similar thing an a few other files.

contrib/brl/bbas/bhdfs/pro/processes/bhdfs_boxm2_load_scene_process.cxx
```diff
@@ -19,8 +19,8 @@
 
 namespace bhdfs_boxm2_load_scene_process_globals
 {
-  const unsigned n_inputs_ = 1;
-  const unsigned n_outputs_ = 1;
+  constexpr unsigned n_inputs_  = 1;
```
dzenanz: The added space is not necessary.

contrib/brl/bbas/bhdfs/pro/processes/bhdfs_fs_manager_processes.cxx
```diff
@@ -12,8 +12,8 @@
 
 namespace bhdfs_create_fs_manager_process_globals
 {
-  const unsigned n_inputs_ = 2;
-  const unsigned n_outputs_ = 1;
+  constexpr unsigned n_inputs_  = 2;
```
dzenanz: The added space thing appears many times.

contrib/rpl/rgrl/tests/test_estimator.cxx
```diff
@@ -63,6 +63,7 @@ namespace {
     return v;
   }
 
+#if 0
   vnl_vector<double>
   random_3d_normal_error( )
```
dzenanz: Is the whole function unused?

core/vnl/vnl_matrix_fixed.h
```diff
@@ -111,6 +111,19 @@ class VNL_TEMPLATE_EXPORT vnl_matrix_fixed
...
   vnl_matrix_fixed() = default;
+  //: Construct an m*n Matrix and copy rhs into it.
+  //  Abort if rhs is not the same size.
+  vnl_matrix_fixed(const vnl_matrix_fixed& rhs) = default;
+  vnl_matrix_fixed(vnl_matrix_fixed&& other) = default;
+  //: Copy another vnl_matrix_fixed<T,m,n> into this.
+  vnl_matrix_fixed& operator=(const vnl_matrix_fixed& rhs) = default;
+  vnl_matrix_fixed& operator=(vnl_matrix_fixed&& rhs) = default;
+  //  Destruct the m*n matrix.
+  // An explicit destructor seems to be necessary, at least for gcc 3.0.0,
```
dzenanz: Comment no longer relevant.

contrib/gel/gevd/gevd_noise.h
hjmjohnson: Fixed:  using =delete to meet comment specification, and the code still passes all tests.

dzenanz: Great!

core/vcsl/vcsl_geocentric.h
hjmjohnson: Fix.  Now explicitly private (instead of default to private).

contrib/brl/bbas/bhdfs/pro/processes/bhdfs_fs_manager_processes.cxx
hjmjohnson: Fixed.

core/vnl/vnl_matrix_fixed.h
hjmjohnson: Done, + many other gcc 3.0 compiler references in comments removed.

core/vcsl/vcsl_geocentric.h
dzenanz: I think there were a few more. Did you update them too?

hjmjohnson: There are 3.    I don't understand the point of this code. 

- If some constructor is private, it means that no one but the class itself (and friends) should be able to create instances of it using that constructor. Therefore, you can provide static methods like getInstance() to create instances of the class or create the instances in some friend class/method.

There is no static or friend class/method.  
 Very confusing, but the code is back to the way it used to be for.  I'll leave it to someone else to fix. this code that I don't need.

dzenanz:
In commit `DOC: Removed comments about unsupported complilers.` there is a misspelling: `complilers`.

core/vcsl/vcsl_geocentric.h
dzenanz: I suspected so. In that case, those constructors could be publicly deleted?

hjmjohnson:
@dzenanz @mleotta @TimCootes @thewtex 

Thanks for the feedback.  If there are no objections, I am going to merge this tomorrow morning.  

NOTE:  I've built with all of contrib on (including BRL) and all the test are passing.  I found 2 old bugs uncovered by removing compiler warnings, those are included here, but not backported any further.

PR #518, opened 2018-10-29
COMP: BRL shared lib inconsistency

### LINKED ISSUE #508: contrib BRL does not build with shared libs
VXL can not be built with BRL turned on due to circular dependencies in the libraries.

 "brad_spectral_angle_mapper::compute_sam_img

 "brad_wv3_bands(std::__1::vector<float, std::__1::allocator<float> >&, std::__1::vector<float, std::__1::allocator<float> >&, int)", referenced from:
      brip_blob_connect_comp_bb(vil_image_view<float> const&, vil_image_view<bool> const&, vil_image_view<bool> const&, float, float, std::__1::vector<unsigned int, std::__1::allocator<unsigned int> >&, std::__1::vector<unsigned int, std::__1::allocator<unsigned int> >&, std::__1::vector<unsigned int, std::__1::allocator<unsigned int> >&, std::__1::vector<unsigned int, std::__1::allocator<unsigned int> >&, std::__1::vector<vgl_polygon<double>, std::__1::allocator<vgl_polygon<double> > >&) in brip_ms_blob_detection.cxx.o

BUILD_GEL not valid with shared libraries due to circular dependancies
+between brad/brip libraries. The brad_wv3_bands and theh brad_spectral_angle_mapper::compute_sam_img
+functions are needed by the brip library, but the the brad library depends upon the brip library."

hjmjohnson:
There were two unrelated problems that caused
shared library building with BRL libraries.

1) Two submodules that were not used external to the unit test for
   those submodules had untested function references (OK for static libs,
   but caused linkage failure for shared libs).
	contrib/brl/bseg/bstm_multi/pro unused/incomplete lib not compiled
        contrib/brl/bseg/bstm_multi/cpp unused/incomplete lib not compiled

2) A circular dependancy where the bbrad and brip libraries depended on each other
   unnecessarily.  By forcing brip to depend on brad only (i.e. removing the dependancy
   from brad to brip) and adding a few missing library dependancies, the
   shared library builds of BRL are working agian.

Resolves #508

### DIFF
--- a/contrib/brl/bbas/brad/CMakeLists.txt
+++ b/contrib/brl/bbas/brad/CMakeLists.txt
@@ -48,7 +48,7 @@
   aux_source_directory(Templates brad_sources)
   vxl_add_library(LIBRARY_NAME brad LIBRARY_SOURCES ${brad_sources} )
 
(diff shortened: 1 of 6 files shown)


hjmjohnson:
Tested with shared libs on for linux and mac with contrib and BRL turned on.

PR #523, opened 2018-10-30
Remove unconditional cxx11 header checks

hjmjohnson:
Many manual checks for various compiler features are not mandated by the C++11 compiler compatibility.


hjmjohnson:
@thewtex @dzenanz FYI:  I already merged this, but I thought you might be interested in how much could be cleaned up (i.e. made irrelevant) with the update to C++11.

thewtex:
Holy :hankey: , nice @hjmjohnson !

PR #524, opened 2018-10-30
Add namespace to cmake options

hjmjohnson:
Provide a more clean interface for selecting options to build.


hjmjohnson:
@thewtex @dzenanz This is the last change I'd like to make to VXL before changing ITK to require at least VXL version 2.0.0 for building.  It hides some of the VXL components from the main screen (mark_as_advanced) and it clearly identifies VXL options as separate from ITK options.

hjmjohnson:
@decrispell Thanks.  My OCD for fixing the VXL code is subsiding :)  I hope that only small changes changes are needed from here on out.   There are some annoying warnings on the dashboard. And there are 2 tests that fail when building absolutely everything (but those two were failing before I started.)

config/cmake/Modules/VXLStandardOptions.cmake
```diff
@@ -32,7 +32,7 @@ if( WIN32 )
 else()
   option( BUILD_SHARED_LIBS "Build shared libraries." OFF)
 endif()
-set( VXL_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS} )
+set( BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS} )
```
dzenanz: This line is no longer necessary.

dzenanz:
Thanks for maintenance!

PR #539, opened 2018-11-08
Revert isnan cxx11 work around

hjmjohnson:


hjmjohnson:
@blowekamp Brad.  Thank you for recognizing this.  Your GCC fix has been restored.

blowekamp:
Great! Thanks for getting it back in!

PR #541, opened 2018-11-13
use binary dir for generated include files, not install dir

decrispell:
Prior to this fix, CMake gave me an error due to trying to write
rrel_export.h.tmp to a system directory (/usr/local/...)


hjmjohnson:
EXCELLENT!  That was really annoying recently.

Issue #526, opened 2018-10-31
BRL unresolved bug.


hjmjohnson:
/Users/johnsonhj/src/vxl/contrib/brl/bbas/brad/brad_spectral_angle_mapper.cxx:354:24: warning: result of comparison of constant 255 with expression of type 'char' is always false
      [-Wtautological-constant-out-of-range-compare]
      if (c <= -1 || c >= 255) { // check if the file contains bad character

hjmjohnson:
Closing as obsolete. The specific -Wtautological-constant-out-of-range-compare warning in brad_spectral_angle_mapper.cxx was addressed in subsequent cleanup work; current builds on master are clean for this warning.

PR #542, opened 2018-11-13
fix tautilogical comparison to silence warning

decrispell:
addresses #526


contrib/brl/bbas/brad/brad_spectral_angle_mapper.cxx
```diff
@@ -351,7 +351,7 @@ void brad_spectral_angle_mapper::add_aster_dir(const std::string aster_dir) {
     std::ifstream is(fi());     // open file
     char c;
```
hjmjohnson: char can be signed or unsigned based on the platform.  I'm OK with the simple way of silencing the warning, but does this test need to deal with different char ranges?

decrispell: Good question, I'm not sure (I'm not familiar with this code).  @andyneff any idea what exactly constitutes an "invalid character" in an aster file?

PR #544, opened 2018-11-13
fix inconsistently defined const members

decrispell:
fixed compile errors by moving ```constexpr``` definitions to header files, replacing the ```const``` declarations there.


contrib/brl/bbas/bdgl/bdgl_curve_algs.h
```diff
@@ -21,9 +21,13 @@
 class bdgl_curve_algs
 {
  public:
-  static bdgl_EXPORT_DATA const double tol;
-  static bdgl_EXPORT_DATA const double max_edgel_sep;
-  static bdgl_EXPORT_DATA const double synthetic;
+  static bdgl_EXPORT_DATA constexpr double tol = 1e-6;
```
hjmjohnson: This was 1e-16 and now is 1e-6.  I think that it may have been a typo.

decrispell: Good catch!  Fixed now.

PR #561, opened 2018-11-28
ENH: compile VXL in parallel on Visual Studio

hjmjohnson:
If VXL is modified, the incremental build becomes sequential during
compilation of VXL.  This enables object-level compile parallelism
in VXL libraries.

### DIFF
--- a/config/cmake/config/vxl_utils.cmake
+++ b/config/cmake/config/vxl_utils.cmake
@@ -92,6 +92,9 @@ macro( vxl_add_library )
   list(LENGTH lib_srcs num_src_files)
   if( ${num_src_files} GREATER 0 )
     add_library(${lib_name} ${lib_srcs} )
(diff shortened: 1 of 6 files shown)


config/cmake/config/vxl_utils.cmake
```diff
@@ -92,6 +92,10 @@ macro( vxl_add_library )
   list(LENGTH lib_srcs num_src_files)
   if( ${num_src_files} GREATER 0 )
     add_library(${lib_name} ${lib_srcs} )
+    # This enables object-level build parallelism in VNL libraries for MSVC
+    if(MSVC)
+      target_compile_definitions(${lib_name} PRIVATE "/MP") # enable object level parallel compilation
```
dzenanz: Since there is a comment on line 95, comment on line 97 can be removed.

PR #568, opened 2018-12-08
ENH: Move definition of vnl_matrix to .hxx

hjmjohnson:
Supporting WINDOWS_EXPORT_ALL_SYMBOLS needed the definition
in one place to avoid a duplicate symbol build failure.

Also remove extra space between comments and code that
artificially separated the desription from the code.


hjmjohnson:
@dzenanz port to upstream.

PR #574, opened 2019-01-14
COMP: Fixes #367: VXLTargets.cmake duplicates

### LINKED ISSUE #367: VXLTargets.cmake duplicates
I'm building a library that uses VXL.
It is complaining of repeated tags in VXLTargets.cmake.
Tested under both Ubuntu 16.04 (cmake 3.5.1) and Mac OS (cmake 3.6.20160910-gd9043)

I can see from vxl/config/cmake/export/VXLCreateProject.cmake
that upon creation of VXLTargets, an APPEND is used:

line 34:    APPEND

Whys is that? Why not overwrite?

I rebuild vxl using VGUI_ON a second time around, and the VXLTargets.cmake keeps getting bigger.

rfabbri:
No reason to append. Just overwrite the file.

(cherry picked from commit lemsvpe@c7de85c33aaaa6a554c01e8b67ae293b87cf18cd)


hjmjohnson:
@dzenanz @thewtex   I'm just linking you here.  This errant "APPEND" was added as part of a large commit that was related to supporting ITK.  I am 99% sure that this change was a mistake, and not needed.

PR #576, opened 2019-02-01
circleci update (c++11, cdash on master only)

drewgilliam:
Update circleci configuration with C+11 requirement.  Also submit circleci build results to open.cdash.org only for master branch (not experimental/development branches).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
             set(CTEST_BINARY_DIRECTORY "/build")
             set(CTEST_COMMAND "/usr/bin/ctest")
             set(CTEST_CONFIGURE_COMMAND "${CMAKE_COMMAND} -DCMAKE_BUILD_TYPE:STRING=\"${CTEST_BUILD_CONFIGURATION}\"")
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -58,9 +58,13 @@ jobs:
           name: Submitting to open.cdash.org
           working_directory: /build
           command: |
-            env | grep ^CIRCLE > circleci.env
-            ctest -A circleci.env -D ExperimentalSubmit -V
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+              env | grep ^CIRCLE > circleci.env
+              ctest -A circleci.env -D ExperimentalSubmit -V
+            else
+              echo "Skipping open.cdash.org submission..."
```
hjmjohnson: Why skip?  We need to see the dashboard for all branch builds.

drewgilliam: Hi Hans!  

There's a few reasons to skip cdash - 
- github.com/vxl/vxl doesn't use circleci, so this doesn't affect anything here
- this step is not run if there was a build error, so cdash would never see any error reports
- personally, I'd like to give users the ability to skip cdash submission on their own forks for very experimental feature sets

I could switch this check to a different circleci environment variable (if the environment variable exists and is true, skip cdash) so users are given the opportunity to disable cdash submission if they so desire?

PR #585, opened 2019-02-11
heightmap_from_disparity adaptor

drewgilliam:
* bpgl_heightmap_from_disparity python adaptor
* volm_combine_height_map_process.cxx: direct threshold for median filter, return "-9999.0" as invalid value
* bsgm_matching_stereo_process.cxx: optional text file output
* vpgl_rectify_images_process accepts dispairty image directly


drewgilliam:
@decrispell - FYI

PR #586, opened 2019-02-13
new vgl functionality

drewgilliam:
New VGL functionality, including:

- casting constructors (e.g., cast vgl_point_2d of type float into type double)
- vgl_intersection (both in vgl and vgl/algo)
- 2D bounding box for vgl_polygon
- small initialization fix for cubic splines
- comment update for clarity


drewgilliam:
@hjmjohnson - cleaned up as requested

PR #594, opened 2019-02-19
vpgl_rational_camera_compute

drewgilliam:
`vpgl_rational_camera_compute::compute` - new vpgl/algo capability to compute a rational camera model from world-to-image correspondences.  Supporting code and unit tests also included.


drewgilliam:
@decrispell - FYI

PR #599, opened 2019-02-27
vpgl_affine_camera

drewgilliam:
Improvements to vpgl_affine_camera
- improve backprojection via direct ray construction
- ray_dir accessor
- postmultiply by translation
- new related tests
- related change to BRL bpgl test_camera_from_box - backproject via ray instead of line_3d_2p


drewgilliam:
@decrispell - FYI

PR #601, opened 2019-02-27
BUG: an absolute path is specified for VCL_COMPILER_DETECTION_HEADER

hjmjohnson:
Problem identified and contributed by @dzenanz from ITK upstream.

### DIFF
--- a/vcl/CMakeLists.txt
+++ b/vcl/CMakeLists.txt
@@ -88,11 +88,11 @@ write_compiler_detection_header(
 # Install platform-specific compiler detection files
 #
 if("${VXL_INSTALL_INCLUDE_DIR}" STREQUAL "include/vxl")
-  set(_compilers_destination ${VXL_INSTALL_INCLUDE_DIR}/vcl/compilers)
-  set(_complier_detection_destination ${VXL_INSTALL_INCLUDE_DIR}/vcl)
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@thewtex @dzenanz

hjmjohnson:
@dzenanz Done.

FYI:
```
cmake -DCMAKE_INSTALL_PREFIX=/tmp/vxl-install ../vxl
make 
make install

===

Then build itk using system vxl against /tmp/vxl-install, and the build succeeded.


```

dzenanz:
The issue does not manifest itself on Linux - I tried on my machine and it worked. But this **doesn't** fix the issue on Windows:
```text
5>-- Installing: C:/Program Files/VXL/bin/vcl.dll
5>-- Up-to-date: C:/Program Files/VXL/include/vxl/vcl/vcl_legacy_aliases.h
5>CMake Error at vcl/cmake_install.cmake:87 (file):
5>  file cannot create directory: C:/Program
5>  Files/VXL/include/vxl/vcl/C:/Misc/vxl-vs2017/vcl.  Maybe need
5>  administrative privileges.
5>Call Stack (most recent call first):
5>  cmake_install.cmake:75 (include)
```
As it is easier to debug this directly in VXL, I suggest we do so.

PR #602, opened 2019-02-27
COMP: Remove test for MIPSPro support by cmake

hjmjohnson:
vcl/tests/test_preprocessor.cxx:37:11: error:
'VXL_COMPILER_IS_MIPSpro' was not declared in this scope

cmake commit beb991110d84e9c0a6b1339e22fa98284b08bac9
  Author: Brad King <brad.king@kitware.com>
  Date:   Fri Jan 11 13:50:46 2019 -0500

  Remove now-unused code once used on IRIX

  We dropped support for IRIX as a host platform long ago.
  Remove some leftover code.

### DIFF
--- a/vcl/tests/test_preprocessor.cxx
+++ b/vcl/tests/test_preprocessor.cxx
@@ -33,8 +33,7 @@ int test_preprocessor_main(int /*argc*/,char* /*argv*/[])
         + VXL_COMPILER_IS_MSVC
         + VXL_COMPILER_IS_ADSP
         + VXL_COMPILER_IS_IAR
-        + VXL_COMPILER_IS_ARMCC
-        + VXL_COMPILER_IS_MIPSpro;
+        + VXL_COMPILER_IS_ARMCC;
 
   int result = 0;
 


hjmjohnson:
Error identified on :  https://open.cdash.org/viewBuildError.php?buildid=5775647

bradking:
It was actually [MR 3004](https://gitlab.kitware.com/cmake/cmake/merge_requests/3004), and that is a regression I need to fix on the CMake side.

hjmjohnson:
@bradking Thanks.  I still think it makes sense to remove this line from the test in VXL.  The required version of cmake for vxl now does not have the ability for that value to be non-zero.

Do you have a suggestion for this patch that would be more appropriate?
 (i.e. also remove VXL_COMPILER_IS_Borland that is also impossible to have a non-zero value with the current version of cmake?)

bradking:
See CMake [MR 3043](https://gitlab.kitware.com/cmake/cmake/merge_requests/3043) for the fix over there.

The patch in this PR for vxl looks fine.  Only the commit message needs to be revised.

PR #604, opened 2019-03-01
BUG: fixing absolute path which slips into install location

dzenanz:
Follow up to #601. Closes InsightSoftwareConsortium/ITK#545.

### DIFF
--- a/vcl/CMakeLists.txt
+++ b/vcl/CMakeLists.txt
@@ -115,8 +115,9 @@ vxl_configure_file(${CMAKE_CURRENT_LIST_DIR}/vcl_where_root_dir.h.in ${PROJECT_B
 
 include_directories(${PROJECT_BINARY_DIR})
 set( vcl_sources
-  vcl_legacy_aliases.h ${VCL_COMPILER_DETECTION_HEADER}
-  vcl_deprecated.cxx  vcl_deprecated.h
+  vcl_legacy_aliases.h
+  vcl_deprecated.cxx
+  vcl_deprecated.h
 
   vcl_cmath.h
   vcl_cassert.h
--- a/vcl/CMakeLists.txt
+++ b/vcl/CMakeLists.txt
@@ -54,7 +54,7 @@ write_compiler_detection_header(
       cxx_generic_lambdas                     # Generic lambdas, as defined in N3649.
       cxx_inheriting_constructors             # Inheriting constructors, as defined in N2540.
       cxx_inline_namespaces                   # Inline namespaces, as defined in N2535.
-      cxx_lambdas                             #Lambda functions, as defined in N2927.
+      cxx_lambdas                             # Lambda functions, as defined in N2927.
       cxx_lambda_init_captures                # Initialized lambda captures, as defined in N3648.
       cxx_local_type_template_args            # Local and unnamed types as template arguments, as defined in N2657.
       cxx_long_long_type                      # long long type, as defined in N1811.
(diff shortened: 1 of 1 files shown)


dzenanz:
@hjmjohnson @thewtex This fixes it for me. Please review and merge. Then ITK's copy can be updated.

vcl/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ write_compiler_detection_header(
       cxx_generic_lambdas                     # Generic lambdas, as defined in N3649.
       cxx_inheriting_constructors             # Inheriting constructors, as defined in N2540.
       cxx_inline_namespaces                   # Inline namespaces, as defined in N2535.
-      cxx_lambdas                             #Lambda functions, as defined in N2927.
+      cxx_lambdas                             # Lambda functions, as defined in N2927.
```
hjmjohnson: Not related.

vcl/CMakeLists.txt
```diff
@@ -65,14 +65,14 @@ write_compiler_detection_header(
       cxx_range_for                           # Range-based for, as defined in N2930.
       cxx_raw_string_literals                 # Raw string literals, as defined in N2442.
       cxx_reference_qualified_functions       # Reference qualified functions, as defined in N2439.
-      cxx_relaxed_constexpr                   #Relaxed constexpr, as defined in N3652.
+      cxx_relaxed_constexpr                   # Relaxed constexpr, as defined in N3652.
```
hjmjohnson: Not related.

vcl/CMakeLists.txt
```diff
@@ -115,8 +115,10 @@ vxl_configure_file(${CMAKE_CURRENT_LIST_DIR}/vcl_where_root_dir.h.in ${PROJECT_B
 
 include_directories(${PROJECT_BINARY_DIR})
 set( vcl_sources
-  vcl_legacy_aliases.h ${VCL_COMPILER_DETECTION_HEADER}
```
hjmjohnson: So is the fix simply replacing ${VCL_COMPILER_DETECTION_HEADER} with the hard coded value vcl_compiler_detection.h ?

Can the variable VCL_COMPILER_DETECTION_HEADER be removed elsewhere?

dzenanz: Yes, that was the fix.

We need `VCL_COMPILER_DETECTION_HEADER` in two places. CMake complains if I remove the full path from that variable.

dzenanz:
I split the commit into bug fix and style change.

@hjmjohnson do we need all of the listed compiler features tested now that we require support for C++11, and VS2015+ on Windows?

hjmjohnson:
@dzenanz There is still a problem with the install phase.

```
CMake Error at vcl/cmake_install.cmake:68 (file):
  file INSTALL cannot find
  "/home/travis/build/vxl/vxl/vcl/vcl_compiler_detection.h".
Call Stack (most recent call first):
  cmake_install.cmake:71 (include)
make: *** [install] Error 1
The command "make install" exited with 2.
```

hjmjohnson:
> I split the commit into bug fix and style change.
> 
> @hjmjohnson do we need all of the listed compiler features tested now that we require support for C++11, and VS2015+ on Windows?

I don't think that they hurt.  They are probably more explicit than is necessary, but that does not hurt.  This was primarily to provide fine grained failure messages for a compiler that does not support a particular feature.

hjmjohnson:
@dzenanz Will you have a look at the dashboard failure?

dzenanz:
If you don't fix the PR before then, I will get around to it sometime next week.

dzenanz:
This works for me now on both Windows and Linux.

dzenanz:
CI is green. Merge?

hjmjohnson:
@dzenanz I'm testing right now.
1) checkout vxl
2) build vxl
3) install vxl
4) build ITK against installed vxl 

We can merge, but I want to ensure that the above sequence also works before merging into ITK.

hjmjohnson:
@dzenanz WORKED ! Thanks for you efforts.  Hans

PR #606, opened 2019-03-11
COMP: Improve header search robustness

hjmjohnson:
The path to headers for the tests should be
consistently relative to the core directory.


hjmjohnson:
Failing test case is not related to this.

PR #608, opened 2019-03-19
Contrib BRL new functions

drewgilliam:
New BRL functionality
- bil_warp: warp an image according to a tomography transform
- bvrml custom color
- bpgl_pointset - from disparity triangulation


drewgilliam:
FYI - @decrispell

PR #609, opened 2019-03-20
vpgl_equi_rectification

drewgilliam:
`vpgl_equi_rectification` - Rectify a pair of images by applying equal and opposite warp transforms to each image, thus reducing scale distortion during rectification.  Includes bugfix for `vpgl_affine_rectification` (vector should be of length 3).

`vpgl_affine_camera_compute` compute viewing distance, additional test


drewgilliam:
@decrispell

PR #614, opened 2019-03-22
COMP: Missing file for install of vxl

hjmjohnson:
The vcl_msvc_warnings.h file was missing from the install
process.  This only affects MSVC installed builds.


hjmjohnson:
@dzenanz @thewtex 
Related to https://github.com/InsightSoftwareConsortium/ITK/issues/602

PR #615, opened 2019-03-28
rational cam print & contrib update

drewgilliam:
CORE VPGL:
- minor change to vpgl_rational_camera print function, increasing precision of output

CONTRIB BRL:
- brip_roi - favor unsigned->int casting (likely safer, as the int values could be negative while the unsigned values are likely to be relatively small)
- adaptor process bugfix - allow rational_cam to have non-integer offset values during crop operation 
- adaptor process - get adaptor verbosity


drewgilliam:
@decrispell

PR #616, opened 2019-03-29
verbosity & contrib adaptors

drewgilliam:
CORE:
- remove print statements from vil_geotiff_header & vpgl_geo_camera (user may print information elsewhere if needed for debugging)

CONTRIB BRL adaptors:
- user-defined verbosity for various adaptor processes
- catch error in bpgl_heightmap_from_disparity_process
- return region list directly in bvgl_geo_index_region_resource_process


drewgilliam:
@decrispell

PR #617, opened 2019-04-03
function was not setting read values

scott-vsi:


scott-vsi:
@decrispell can you review this

PR #620, opened 2019-04-09
vgl_infinite_line_3d constructor fix

scottsorensenvision:
Fixing a constructor that did not copy all fields and caused issues with line direction. Added a test case that would fail without the fix.

### DIFF
--- a/core/vgl/vgl_infinite_line_3d.hxx
+++ b/core/vgl/vgl_infinite_line_3d.hxx
@@ -11,14 +11,9 @@
 #  include <vcl_msvc_warnings.h>
 #endif
 template <class Type>
-vgl_infinite_line_3d<Type>::vgl_infinite_line_3d(vgl_point_3d<Type> const& p1,
-                                                 vgl_point_3d<Type> const& p2)
-{
-  vgl_vector_3d<Type> dir = p2-p1;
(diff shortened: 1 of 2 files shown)


scottsorensenvision:
@decrispell Could you please review this?

core/vgl/vgl_infinite_line_3d.hxx
```diff
@@ -16,8 +16,7 @@ vgl_infinite_line_3d<Type>::vgl_infinite_line_3d(vgl_point_3d<Type> const& p1,
 {
   vgl_vector_3d<Type> dir = p2-p1;
   vgl_infinite_line_3d<Type> l(p1, dir);
-  x0_ = l.x0();
-  t_ = dir;
+  *this = l;
```
decrispell: Would a delegating constructor be a more idiomatic way of expressing this?:
```
vgl_infinite_line_3d<Type>::vgl_infinite_line_3d(vgl_point_3d<Type> const& p1, vgl_point_3d<Type> const& p2)
: vgl_infinite_line_3d<Type>(p1, p2-p1)
{}
```

scottsorensenvision: I have updated the PR with your suggestion.

PR #622, opened 2019-04-10
Move vpgl_lvcs destructor to cxx file

drewgilliam:
vpgl_lvcs is derived from vbl_ref_count, thus the destructor is implicitly virtual. As the destructor is the only virtual function in vpgl_lvcs, DO NOT define inline; the destructor must be the "key function" (first non-inline, non-pure, virtual function).

Defining in cxx protects against "undefined reference to vtable" linking errors.
https://gcc.gnu.org/wiki/VerboseDiagnostics#missing_vtable


drewgilliam:
@decrispell

PR #631, opened 2019-04-25
CONTRIB BRL: update isfm python adaptors

drewgilliam:
update BRL python isfm adaptor processes
- make isfm adaptor process accept 3D region directly
- eliminate requirement for ASTER DEM ingestion
- general code cleanup (e.g., removing #if statements)


drewgilliam:
@decrispell

PR #633, opened 2019-05-03
CONTRIB BRL: affine camera calibration

drewgilliam:
New BRL submodule to perform calibration of multiple affine cameras (typically derived from rational cameras), developed by @mundyj.


drewgilliam:
@decrispell

PR #634, opened 2019-05-03
CONTRIB OpenCL

drewgilliam:
Several BRL packages use OpenCL - refactor to use find_package
- OpenCL find_package (see [here](https://cmake.org/cmake/help/v3.10/module/FindOpenCL.html))
- update cmake variables for find_package compatibility
- Consolidate OpenCL discovery in BRL cmake
- delete FindOpenCL.cmake from config directory
- add CL_TARGET_OPENCL_VERSION preprocessor value (see [here](https://github.com/KhronosGroup/OpenCL-Headers#compiling-for-a-specific-opencl-version))


drewgilliam:
@decrispell & @andyneff  - I think I got all the OpenCL stuff right, can you please take a look-see?  Also let me know if you'd like to target a different OpenCL version by default.

https://github.com/KhronosGroup/OpenCL-Headers/blob/de26592167b9fdea503885e40e8755393c56523d/CL/cl_version.h#L33

PR #641, opened 2019-05-24
CONTRIB BRL: minor acal update

drewgilliam:
CONTRIB BRL minor update for acal
-- added all match tree accessor to match_graph
-- made solution error its own class
-- minor code cleanup


drewgilliam:
@decrispell

PR #642, opened 2019-05-24
CONTRIB BRL : bsgm update

drewgilliam:
CONTRIB BRL update for bsgm (semi-global matching)
- move error checking to separate file
- new left-right consistency check
- optional directional bias
- additional debugging options & code cleanup


drewgilliam:
@decrispell

PR #646, opened 2019-06-03
ENH: vpgl_lvcs const correctness

drewgilliam:
## PR Checklist
🚫 Makes breaking changes to vxl/core API that requires semantic versioning increase
✅ Makes design changes to existing vxl/core API that requires semantic versioning increase
🚫 Adds tests and baseline comparison (quantitative)
🚫 Adds Documentation

## PR Description
- Update `vpgl_lvcs` functions for const correctness (ensure all functions which do not change the state of the object instance are const)
- Update `vpgl_lvcs` protected member functions (change name from `set_*_conversions` to `get_*_conversions` and use protected member variables directly)
- Eliminate (now) unnecessary `const_cast` from `vpgl_local_rational_camera`
- tweak version number - now v2.0.2.1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ foreach(p
 endforeach()
 
 project(VXL #Project name must be all caps to have properly generated VXL_VERSION_* variables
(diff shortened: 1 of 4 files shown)


drewgilliam:
@decrispell

PR #647, opened 2019-06-06
ENH: contrib/brl python adaptor infrastructure

drewgilliam:
## PR Checklist
:no_entry_sign: Makes design changes to existing vxl/core\* API that requires semantic versioning increase
:no_entry_sign: Makes changes to the contributed directory API DOES NOT require semantic versioning increase
:no_entry_sign: Adds tests and baseline comparison (quantitative).

## PR Description
Update contrib/brl python adaptor infrastructure:
- eliminate local FindPython.cmake, instead using the standard cmake find_package(PythonLibs)
- update adaptors for python3 compatibility
- note adaptors are still fully compatible with python 2.7 - users who need control of the python version see Python_ADDITIONAL_VERSIONS discussed [here](https://cmake.org/cmake/help/v3.10/module/FindPythonLibs.html)


drewgilliam:
@decrispell

PR #650, opened 2019-06-13
ENH: contrib/brl/bbas/bres loader

NoahRJohnson:
## PR Checklist

## PR Description
Created a BRL resource loader, allowing resource files in the source code to be found at runtime. The loader will search $VXL_DATA_DIR, $VXL_ROOT_SOURCE_DIR, and $CMAKE_INSTALL_FULL_DATAROOTDIR in that order.

This allows our docker images to avoid mounting in the VXL build directory.


hjmjohnson:
Can this be simplified to make it easier to review?  There are conflicts that need to be addressed.

NoahRJohnson:
@hjmjohnson Thanks for taking a look.

I've rebased onto master, that should have fixed the conflicts. What do you mean by simplifying the PR? I could squash the commits.

hjmjohnson:
The current patches with names like "Fixed bugs" and "renamed" are not that helpful.  

Each patch should be 1 topic (i.e. a self contained story of a single change).   

It's hard to review the current topic because a single logical change in the PR is spread across several patches.  If I comment on an error in the first patch, the second or third patch may change the code from the first patch.  If you want the end result of the entire PR to be evaluated, then just squash it into one omnibus change for reviewing with a single omnibus commit message that describes all the changes made.

My approach to making good patches is to do the following.

```
git reset origin/master   #  have all of the changes relative to origin/master not staged
git add -p                        # Review each code change, and add all one topic changes to staging area
git commit                       # Commit a message for this topic
git add -p                        # Review code changes for next topic.
git commit                       # Commit next topic changes
```

The advantage of the is approach is that each patch can be independently reviewed, independently  reverted.

A second advantage of this approach is that your final commits make it look like you never make a mistake!  Only perfect code refactorings exist in your commits!

PR #654, opened 2019-07-09
PERF/ENH: BSGM memory, BPGL rectification

drewgilliam:
## PR Checklist
:no_entry_sign: Makes design changes to existing vxl/core/\* API that requires semantic versioning increase
:no_entry_sign: Makes changes to the contributed directory API DOES NOT require semantic versioning increase
:no_entry_sign: Adds tests and baseline comparison (quantitative).

## PR Description
Contrib changes as follows:
- BSGM: 2x reduction in `bsgm_disparity_estimator` memory consumption  (eliminating separate census and x-gradient matrices, favoring only a single fused matrix) + additional census cost refactoring 
- BGPL: allow user to define invalid pixel fill value for `bpgl_rectify_affine_image_pair`


drewgilliam:
@decrispell

PR #656, opened 2019-07-15
BUG: revert std::stod/stoi for c-style code

drewgilliam:
## PR Checklist
:no_entry_sign: Makes design changes to existing vxl/core\* API that requires semantic versioning 
:no_entry_sign: Makes changes to the contributed directory API DOES NOT require semantic versioning 
:no_entry_sign: Adds tests and baseline comparison (quantitative)

## PR Description
PR #653 converted atoi -> std::stoi and atof -> std::stod.  This change breaks c-style libraries which are unable to access c++ std library features.   Revert c-style libraries to atoi and atof.  

Note this issue was not caught as VXL tests do not build contrib.

### DIFF
--- a/contrib/brl/b3p/shapelib/dbfopen.c
+++ b/contrib/brl/b3p/shapelib/dbfopen.c
@@ -109,13 +109,13 @@
  * Avoid crashing if field or record are out of range in dbfread*attribute().
  *
  * Revision 1.22  1999/12/15 13:47:24  warmerda
- * Added stdlib.h to ensure that std::stod() is prototyped.
+ * Added stdlib.h to ensure that atof() is prototyped.
  *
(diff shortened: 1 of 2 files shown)


drewgilliam:
@hjmjohnson

drewgilliam:
@hjmjohnson - the files are in v3p and b3p (the BRL version of v3p) which in my understanding consist of 3rd party libraries distributed with VXL.  It looks like we don't mess with those 3rd party libraries too much - v3p/tiff, v3p/png, v3p/jpeg, and more are all still c-style.  Conversion of 3rd party libraries to c++ vs. just fixing this compilation bug sounds like a larger issue better handled on the issues page?

hjmjohnson:
@drewgilliam Excellent points.  This reversion is probably the best option.

PR #658, opened 2019-07-17
COMP: No longer disable obsolete MSVC warning C4231 in VNL/vxl/vcl

hjmjohnson:


hjmjohnson:
@dzenanz

N-Dekker:
Thanks Hans! (Off-topic) Could you possibly still have a look at https://github.com/vxl/vxl/blob/master/vcl/vcl_compiler.h.in#L202 ?

    #ifdef VNL_CONFIG_LEGACY_METHODS

It seems to me that it should be an `#if`, instead of an `#ifdef`. At least, in ITK! See my comment at https://github.com/InsightSoftwareConsortium/ITK/pull/1084#issuecomment-510976931

Of course, it's just a matter of how to define `VNL_CONFIG_LEGACY_METHODS` (or not)!

PR #661, opened 2019-08-14
COMP: Improve default building of VXL with MSVC 1920-2929 version

hjmjohnson:
See https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes?view=vs-2019

and discussions:

https://discourse.itk.org/t/has-anyone-tried-to-build-visual-studio-c-2019/1561


hjmjohnson:
@dzenanz  I will test this with ITK soon.

PR #662, opened 2019-08-17
BUG: Fix issue #663, make vnl_bignum constructor for int types generic

N-Dekker:
Made `vnl_bignum` constructor for integer types (`long`, `unsigned long`,
`int`, `unsigned int`) generic, reducing duplicate code and supporting
other integer types as well (including `long long` and `size_t`).

Fixed issue #663 (Undefined behavior vnl_bignum(INT_MIN) and
vnl_bignum(LONG_MIN)).

Added `vnl_bignum(LONG_MAX)` and `vnl_bignum(LONG_MIN)` tests to
test_bignum `run_constructor_tests()`, and added new "to_string" tests,
`run_to_string_tests()`, ensuring that issue #663 is indeed fixed.


N-Dekker:
The following line of code appears troublesome when `l` is `LONG_MIN`:

    l = -l;                     // Get absolute value of l

At  https://github.com/vxl/vxl/blob/v2.0.2/core/vnl/vnl_bignum.cxx#L31 

And there's a similar problem when `l` is `INT_MIN`', at https://github.com/vxl/vxl/blob/v2.0.2/core/vnl/vnl_bignum.cxx#L55

hjmjohnson:
@N-Dekker Can this be fixed with:

https://en.cppreference.com/w/cpp/numeric/math/signbit
https://en.cppreference.com/w/cpp/numeric/math/fabs

N-Dekker:
> Can this be fixed with:
> https://en.cppreference.com/w/cpp/numeric/math/signbit
> https://en.cppreference.com/w/cpp/numeric/math/fabs

@hjmjohnson Thanks Hans, but it that a suggestion or a question?

I do have a fix in mind, but I would first like to be sure that:

1.  this problem can be reproduced at Travis CI
2.  it is a new (unknown) issue that `vnl_bignum(INT_MIN)` and `vnl_bignum(LONG_MIN)` produce incorrect results
3.  people like to have it fixed

hjmjohnson:
@N-Dekker it is a suggestion. Now that we have C++11 support, perhaps there is a more modern/less convoluted way to address this issue.  

Thanks for demonstrating the problem. 

It would be GREAT to have a fix.

N-Dekker:
Please wait... another amend coming soon!

core/vnl/tests/test_bignum.cxx
```diff
@@ -820,6 +867,20 @@ static void run_shift_tests()
 #endif
 }
 
+static void run_to_string_tests()
+{
+  test_min_and_max_and_decimal_digits_to_string<signed char>();
+  test_min_and_max_and_decimal_digits_to_string<unsigned char>();
```
hjmjohnson: Nice Testing!

Issue #664, opened 2019-08-17
vnl_rational should have same size on all platforms


N-Dekker:
### Description

The numerator and the denominator of `vnl_rational` are defined as `long` integers: https://github.com/vxl/vxl/blob/v2.0.2/core/vnl/vnl_rational.h#L76

On Linux, `long` is usually 64 bits, whereas on Windows (using MSVC), a `long` is typically only 32 bits. This may lead to different results, when a calculation that depend on `vnl_rational` is performed both on Windows and Linux.

### Steps to Reproduce

    TEST("(1/INT_MAX)/2 should be greater than zero",
        vnl_rational{ INT_MAX }.invert()/2 > 0, true)

### Expected behavior

Test should pass successfully.

### Actual behavior

On Windows (VS2017 x64) the test will fail, because on Windows `vnl_rational` cannot represent any positive rational number that is less than 1/INT_MAX.

### Reproducibility

100%

### Versions

Tag v2.0.2, as well as the current master, revision aaf7a60f34e7f5e12272a431eed860e02f2d84fa
### Environment

Windows, VS2017 x64

### Additional Information

Instead of `long`, I would suggest using a C++11 fixed width integer type, `std::int64_t`, for both the numerator and the denominator. Such a change would not affect Linux users, but it would fix the issue on Windows platforms, as far as I can see.

N-Dekker:
Fixed by pull request #665 (ENH: Fix issue #664: vnl_rational numerator, denominator both int64_t), merged to the master branch by Hans @hjmjohnson commit 9f4685c03a5c935993bc09a8517415f2d6ce5330

PR #665, opened 2019-08-20
ENH: Fix issue #664: vnl_rational numerator, denominator both int64_t

N-Dekker:
Fixed issue #664, "vnl_rational should have same size on all platforms":

Added `vnl_rational::int_type`, the integer type of both the numerator
and the denominator, and defined this type as `std::int64_t` to ensure
64-bit integers on all supported platforms. Note that `boost::rational`
also has an `int_type`: www.boost.org/doc/libs/1_71_0/boost/rational.hpp

Extensively used C++ SFINAE (Substitution Failure Is Not An Error) to
avoid redundant code.

Added missing `const` keyword to converting operator for signed
integral types.

Added `test_converting_constructors()` to test_rational.

### DIFF
--- a/core/vnl/tests/test_rational.cxx
+++ b/core/vnl/tests/test_rational.cxx
@@ -1,6 +1,9 @@
 #include <iostream>
 #include <iomanip>
 #include <complex>
+#include <string>
+#include <utility>
+#include <typeinfo>
 #include <vnl/vnl_rational.h>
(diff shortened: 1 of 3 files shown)


N-Dekker:
@hjmjohnson Please review 😃 Most changes are quite straightforward, `long` -> `int_type`, I only find the SFINAE a bit tricky, but you'll get used to it 😄

PR #672, opened 2019-09-10
Fix vgui range map warning c4804

clucock1:
COMP: warning C4804: '-': unsafe use of type 'bool' in operation
COMP: warning C4805: '!=': unsafe mix of type 'int' and type 'bool' in operation
COMP: warning C4146: unary minus operator applied to unsigned type, result still unsigned


hjmjohnson:
@clucock1 We do not merge "WIP:" patches.  Please change the name, or squash with the other PR.

Each patch set should be a complete code changes (i.e. not a WIP).

Thanks.

PR #683, opened 2019-09-26
PERF: Add C++11 move-constructor and move-assignment to vnl_vector

N-Dekker:
Added fast (O(1)), `noexcept` move-constructor and move-assignment
operator to `vnl_vector`.

Added `noexcept` specifier to `vnl_vector::swap`, in order to use it
for the move-assignment implementation.

Added `vnl_vector_test_move_construct()` and `vnl_vector_test_move_assign()`.

### DIFF
--- a/core/vnl/tests/test_vector.cxx
+++ b/core/vnl/tests/test_vector.cxx
@@ -9,6 +9,9 @@
 #include <vnl/vnl_cross.h>
 #include <testlib/testlib_test.h>
 
+#include <numeric> // For iota.
+#include <utility> // For move.
+
 void vnl_vector_test_int()
 {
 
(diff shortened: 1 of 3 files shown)


N-Dekker:
@hjmjohnson 
> GREAT TESTS!!! Thanks, 1,000,000 times over for attacking these performance benefits!

You're welcome, Hans! However, I feel there might still be a bug in my code. I have to double-check! That's why it's now temporarily "WIP"!

hjmjohnson:
Please rebase on the master branch for the next submission and testing.

N-Dekker:
My original move-assignment implementation was buggy, but it's fixed now. Did you also spot the bug?
https://github.com/vxl/vxl/blob/1e7285033c7907c788ba79ece56ec16ef7f343ca/core/vnl/vnl_vector.hxx#L373-L382

Anyway, I think it's fixed now, by implementing move-assignment in terms of move-construct + swap, instead. Please check again!

PR #688, opened 2019-09-30
BUG: disable travis cache

drewgilliam:
Incremental travis build appears to be masking full build errors (possibly due to PR #683).  Temporarily disable incremental build.
 
## PR Checklist
🚫 Makes design changes to existing vxl/core\* API that requires semantic versioning increase
🚫 Adds tests and baseline comparison (quantitative).
🚫 Adds Documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,11 @@ env:
 git:
   depth: 3
 
-# cache the build dir
-cache:
-  timeout: 1000
-  directories:
-  - ${PROJ_BUILD_DIR}
+# # cache the build dir
+# cache:
+#   timeout: 1000
+#   directories:
+#   - ${PROJ_BUILD_DIR}
 
 # general dependencies
 addons:
--- a/core/vnl/tests/test_vector.cxx
+++ b/core/vnl/tests/test_vector.cxx
@@ -9,9 +9,6 @@
 #include <vnl/vnl_cross.h>
 #include <testlib/testlib_test.h>
 
-#include <numeric> // For iota.
-#include <utility> // For move.
-
 void vnl_vector_test_int()
 {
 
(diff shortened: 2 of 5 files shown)


drewgilliam:
@hjmjohnson & @N-Dekker - see https://travis-ci.org/drewgilliam/vxl/builds/591617133

This is a full rebuild based on the commit referenced in PR #683, and seems to introduce a few VNL related failures.

N-Dekker:
Thanks for reporting this issue, @drewgilliam I'm not sure what's going wrong here. The move-constructor and move-assignment that I wrote look entirely fine to me. However, it might be that there's a conflict with the logic of the old `vnl_vector_ref`.  I'm not sure. Is it easy for you to reproduce the issue?

drewgilliam:
@hjmjohnson - FYI, this change performs a full build (ignoring any Travis cache) and I don’t expect it to successfully pass Travis until PR #683 is fixed or reverted (and the new master is merged into this branch).

@N-Dekker - a full core build always reproduces the error, see the linked build report above to see what tests are failing.  I frankly don’t understand the issue, so I’m not of much help other than asking to revert the PR.

N-Dekker:
@drewgilliam I'm currently having a closer look at what's going on. Please wait!

hjmjohnson:
@N-Dekker I soo am looking into this, and it does appear like the failures are related to the vnl_vector_ref.


It also appears that the assignment operator is not checking for self-assignment.

N-Dekker:
> It also appears that the assignment operator is not checking for self-assignment.

Thanks Hans, but it's not always required for an assignment-operator to check for self-assignment. As long as the result of the assignment is still valid, of course! You may add self-assignment checks, of course, but I don't think that solves the crash. Please look at my example code again:

    vnl_vector_fixed<int, 3> fixedVector(42);
    std::vector<vnl_vector<int>> stdVector;
    stdVector.push_back(fixedVector.as_ref());

When the `vnl_vector_ref` is pushed-back into the `std::vector<vnl_vector<int>>`, it looses the information that the data is actually from a `vnl_vector_ref`. The `std::vector` treats it as a regular `vnl_vector`, and when it is destructed, `vnl_vector::data` is deallocated. Right? But this `vnl_vector::data` is actually the data of a `vnl_vector_fixed`! Which should not be deallocated at all!!

I think it would be worth reconsidering the design of  `vnl_vector_ref`...

hjmjohnson:
@N-Dekker Thank you for looking into this.  I do think that vnl_vector_ref, and it's design may not be compatible with the move constructors.

I don't have much more time to dedicate to this.  

So that we are on the same page:
   1.  Will you please make a PR with a `git revert 3eb3de2717ddc164e5627214e2ef37d3735d401` so that the master branch is cleanly building/testing? 
  2. Will you then do a `git revert (reverted hash)` and start a new PR that has the previous code + other fixes necessary as a WIP?
   3.  I would like to leave this in your hands and move my attention back to some student requests that I have outstanding.

drewgilliam:
Changes already merged as part of PR #691.  Closing this PR.

PR #692, opened 2019-10-01
Revert "PERF: Add C++11 move-constructor and move-assignment to vnl_vector"

drewgilliam:
Reverts vxl/vxl#683


drewgilliam:
revert request as per @hjmjohnson 
https://github.com/vxl/vxl/pull/688#issuecomment-537169063

FYI @N-Dekker

PR #693, opened 2019-10-01
WIP: PERF: Add C++11 move-constructor and move-assignment to vnl_vector

drewgilliam:
Added fast (O(1)), `noexcept` move-constructor and move-assignment
operator to `vnl_vector`.

Added `noexcept` specifier to `vnl_vector::swap`, in order to use it
for the move-assignment implementation.

Added `vnl_vector_test_move_construct()` and `vnl_vector_test_move_assign()`.

### DIFF
--- a/core/vnl/tests/test_vector.cxx
+++ b/core/vnl/tests/test_vector.cxx
@@ -9,6 +9,9 @@
 #include <vnl/vnl_cross.h>
 #include <testlib/testlib_test.h>
 
+#include <numeric> // For iota.
+#include <utility> // For move.
+
 void vnl_vector_test_int()
 {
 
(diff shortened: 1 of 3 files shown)


drewgilliam:
WIP created in behalf of @N-Dekker, as requested by @hjmjohnson 

https://github.com/vxl/vxl/pull/688#issuecomment-537124703
https://github.com/vxl/vxl/pull/688#issuecomment-537159320

N-Dekker:
In order to prevent crashes previously reported by @drewgilliam and discussed at https://github.com/vxl/vxl/pull/688, it would help if the data members of `vnl_vector` would be made `private`, instead of `protected`. This would stop a derived class, like `vnl_vector_ref`, from interfering with the memory allocation/deallocation of `vnl_vector`.

`vnl_vector_ref` could then be adapted, or replaced by a class like `std::span` (C++20) https://en.cppreference.com/w/cpp/container/span

hjmjohnson:
work superceeded by #694

PR #694, opened 2019-10-02
ENH: Create a working move constructor and assignment operator for vnl_vector and vnl_matrix

hjmjohnson:
STYLE:  Simplify code by removing manual RTO optimizations
    
    RTO optimizations have been a part of the C++ language for
    a long time, and are rubustly implemented by compilers.
    
    Greatly simplify the code readability by removing
    the now unnecessary manual RTO implementations.

See: https://shaharmike.com/cpp/rvo/


hjmjohnson:
@N-Dekker These code changes are made in an attempt to simplify the codebase and make it easier to identify what needs to be done for move constructors.

hjmjohnson:
@dzenanz Would you mind reviewing this PR for vxl?

core/vnl/vnl_vector.h
```diff
@@ -214,14 +198,65 @@ class VNL_EXPORT vnl_vector
...
 
-  vnl_vector<T> operator+(T v) const { return vnl_vector<T>(*this, v, vnl_tag_add()); }
-  vnl_vector<T> operator-(T v) const { return vnl_vector<T>(*this, v, vnl_tag_sub()); }
-  vnl_vector<T> operator*(T v) const { return vnl_vector<T>(*this, v, vnl_tag_mul()); }
-  vnl_vector<T> operator/(T v) const { return vnl_vector<T>(*this, v, vnl_tag_div()); }
+  vnl_vector<T> operator+(T v) const {
+    vnl_vector<T> result(this->size());
+    std::transform(this->begin(), this->end(), result.begin(),
+                   [v](T d) -> T { return d + v; });
+    return result;
```
N-Dekker: `return result` is potentially expensive, because `vnl_vector` does not have a no fast move-constructor. I think it would be better to postpone this change until we have restored fast move-semantics for `vnl_vector`. (Because I think you cannot fully rely on NRVO -- named return value optimization.)

core/vnl/vnl_matrix.h
```diff
@@ -300,23 +282,114 @@ class VNL_EXPORT vnl_matrix
...
-  vnl_matrix<T> operator+(T const& v) const { return vnl_matrix<T>(*this, v, vnl_tag_add()); }
+  vnl_matrix<T> operator+(T const& v) const {
+    vnl_matrix<T> result(this->rows(), this->cols());
+    const unsigned int n = this->num_rows * this->num_cols;
+    T const *m = this->data[0];
+    T *dst = result.data[0];
+
+    for (unsigned int i = 0; i < n; ++i)
+      dst[i] = T(m[i] + v);
+    return result;
```
N-Dekker: `return result` is potentially expensive, because `vnl_matrix` does not have a fast move-constructor. I think it would be better to postpone this change until we have fast move-semantics for `vnl_matrix`. (Because I think you cannot fully rely on NRVO -- named return value optimization.)

core/vnl/vnl_vector.h
```diff
@@ -19,7 +19,7 @@
 #ifdef _MSC_VER
 #  include <vcl_msvc_warnings.h>
 #endif
-#include <vnl/vnl_tag.h>
+
```
dzenanz: Commit message: R**T**O -> R**V**O
r**u**bustly -> r**o**bustly

core/vnl/vnl_vector.h
```diff
@@ -439,33 +474,42 @@ ::put(size_t i, T const& v)
...
 
 //: subtract vector from scalar. O(n).
 // \relatesalso vnl_vector
 template<class T>
 inline vnl_vector<T> operator-(T s, vnl_vector<T> const& v)
 {
-  return vnl_vector<T>(-v, s, vnl_tag_add());
+  vnl_vector<T> result(v.size());
+  for(size_t i=0; i< result.size(); ++i)
+    result[i]= s-1*v[i];
```
dzenanz: Why is `1` required here? Isn't `result[i] = s-v[i];` sufficient?

N-Dekker:
> These code changes are made in an attempt to simplify the codebase and make it easier to identify what needs to be done for move constructors.

@hjmjohnson While I do appreciate your attempt to simplify the codebase, it appears to me that the real problem that prevents proper (fast) implementation of move-semantics for `vnl_vector`  is that its data isn't `private`, and its memory allocation/deallocation can be overruled by derived classes like `vnl_vector_ref`.

Personally I find `vnl_vector_ref` quite complicated, and I hope it can be replaced by something as simple as `std::span` (C++20).

core/vnl/vnl_vector.hxx
```diff
@@ -106,96 +119,68 @@ vnl_vector<T>::vnl_vector (vnl_vector<T> const& v)
...
-  for (size_t i=0; i<num_elmts; ++i)
-    data[i] = u[i] - v[i];
+  // Copy the data pointer and its length from the source object.
+  this->operator=(std::move(rhs));
 }
 
+//: Move-assigns rhs vector into lhs vector. O(1).
 template<class T>
-vnl_vector<T>::vnl_vector (vnl_vector<T> const &u, T s, vnl_tag_mul)
+vnl_vector<T>& vnl_vector<T>::operator=(vnl_vector<T>&& rhs) noexcept
```
N-Dekker: Of course, I feel somewhat sad that the move-assignment-operator that I proposed with pull request #693 did not make it. While it was so simple and elegant, just doing a move-construct and a swap.  😢 This one you're proposing here still does dynamic memory allocation, and can still possibly fail, even though it says `noexcept`. I see why you're proposing it this way, but still it looks a bit troublesome.

N-Dekker:
As an alternative approach, maybe an extra template parameter `SizeSpecifier` could be added to `vnl_vector<T>`, to specify whether the size is fixed or dynamic. By default, it would be dynamic, just like the original `vnl_vector` implementation:

    template <typename T, typename SizeSpecifier = vnl_dynamic_size> class vnl_vector;

But it could be specialized for fixed size:

    template <typename T, std::size_t N>
    class vnl_vector<T, vnl_fixed_size<N>>
    {
        // Implemented just like vnl_vector_fixed<T, N>.
    };

`vnl_dynamic_size` and `vnl_fixed_size<N>` would just be "tags" (empty structs).

Generic functions could then use `vnl_vector<T, SizeSpecifier>`, to support both dynamic and fixed size vectors. For convenience, `vnl_vector_fixed<T>` would be defined as alias of `vnl_vector<T, vnl_fixed_size<N>>`. What do you think?

core/vnl/vnl_vector.hxx
hjmjohnson: I agree 100%, I view this as a stepping stone move forward.  Removing the reliance on memory allocation in the move constructor is also possible (I think), but we need to stop the willy-nilly implied conversion of types through a vnl_vector_ref that previously always resulted in a memory copy from the assignment operator.  For the moment, this 'hack' in the move operator is to overcome the other design discrepancies.

I think that making all the single argument copy constructors 'explicit' is the next chore for avoiding move-constructor problems.

Here is the crux of the problem that requires memory allocation from from the move assignment.   There are many improvements that need to be done, but my feeling this that this solution is "no worse" than the current situation from a security standpoint, and it provides performance benefits for a large number of instances.

```c++
// psuedocode
{
double mutable_contig_memory[4]{1,2,3,4};
vnl_vector_ref<double> myref( mutable_contig_memory, 4 );
// vnl_vector_ref is-a vnl_vector

vnl_vector<double> M;

M.operator=(  vnl_vector_ref<double>( mutable_contig_memory, 4 ) );
//                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This is an rvalue!
assert(M[3] == 4)
M[3] = 99;
assert(M[3] == 99)
assert(mutable_contig_memory[3] == 99 ); // <- what should the answer be?
assert(mutable_contig_memory[3] == 4 );    // <- what should the answer be?

M.set_size(1000); // <- This will attempt to delete the memory
M.fill(7);
assert(M[77] == 7)
}

{
constexpr double immutable_contig_memory[4]{1,2,3,4};
vnl_vector_ref<double> myref( immutable_contig_memory, 4 );
// vnl_vector_ref is-a vnl_vector

vnl_vector<double> M;

M.operator=(  vnl_vector_ref<double>( immutable_contig_memory, 4 ) );
//                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This is an rvalue!
assert(M[3] == 4)
M[3] = 99;
assert(M[3] == 99)
// Answer is clear hear: assert(immutable_contig_memory[3] == 99 ); // <- what should the answer be?
assert(immutable_contig_memory[3] == 4 );    // <- what should the answer be?

M.set_size(1000); // <- This will attempt to delete the memory
M.fill(7);
assert(M[77] == 7)
}
```

I really appreciate your feedback.  The current implementation is focused on identifying the scenarios that need to be supported, and then working on intermediate "it-works and passes all tests" solutions that attempt to address only a few issues at a time.

My goal is that this long road will result in your solution after the other deficiencies are addressed.

N-Dekker: @hjmjohnson I think your code example shows that the current `vnl_vector_ref` logic is very fragile. Right? However, I still think a move-constructor and a move-assignment operator should only be marked `noexcept` if they never fail. Otherwise, I think it's better to remove the `noexcept` keywords from these two member functions.

What do you think of my suggestion to add an extra `SizeSpecifier` template parameter to `vnl_vector<T>`, as an alternative way to allow generic code that supports both fixed and dynamic vnl-vectors? https://github.com/vxl/vxl/pull/694#issuecomment-538502074 Do you think that would be a feasible approach, given the current code base?

hjmjohnson:
> As an alternative approach, maybe an extra template parameter `SizeSpecifier` could be added to `vnl_vector<T>`, to specify whether the size is fixed or dynamic. By default, it would be dynamic, just like the original `vnl_vector` implementation:
> 
> ```
> template <typename T, typename SizeSpecifier = vnl_dynamic_size> class vnl_vector;
> ```
> 
> But it could be specialized for fixed size:
> 
> ```
> template <typename T, std::size_t N>
> class vnl_vector<T, vnl_fixed_size<N>>
> {
>     // Implemented just like the original vnl_vector_fixed<T, N>.
> };
> ```
> 
> `vnl_dynamic_size` and `vnl_fixed_size<N>` would just be "tags" (empty structs).
> 
> Generic functions could then use `vnl_vector<T, SizeSpecifier>`, to support both dynamic and fixed size vectors. For convenience, `vnl_vector_fixed<T, N>` would be defined as alias of `vnl_vector<T, vnl_fixed_size<N>>`. What do you think?

I think this is a great approach, but it is far far far beyond what I can contribute to.  My interest in vxl stops at it's use in ITK.

hjmjohnson:
@N-Dekker the more I think about your `class vnl_vector<T, vnl_fixed_size<N>>` solution, the more I think it is a good approach.  Is this a task that you would be willing to undertake?

hjmjohnson:
Verified that no new valgrind errors exist!  Yeah!   This is stepping stone refactoring that setup up consistent code for the next set of refactorings that would hopefully provide performance gains.

The current vnl_[vector|matrix]_ref classes require a compromise on the move constructor/assignments that limit the potential optimizations.   These should be 'no_except', but currently, they may be required to allocate memory which may throw an exception.  

ITK as been compiled against this version of VXL as well.

Performance testing of the average of 5 runs of the tests resulted in this version and the original master branch version being essentially identical.

core/vnl/vnl_vector.hxx
hjmjohnson: Yes,  very very fragile.  I don't like it at all.  I like your suggestion a lot, but I don't have the resources or time to take on that project all at once.

hjmjohnson:
@dzenanz A lot more work was done with this over the weekend.  Would you mind taking a second look?

core/vnl/vnl_matrix_fixed.h
```diff
@@ -114,10 +114,15 @@ class VNL_EXPORT vnl_matrix_fixed
   //: Construct an m*n Matrix and copy rhs into it.
   //  Abort if rhs is not the same size.
   vnl_matrix_fixed(const vnl_matrix_fixed& rhs) = default;
-  vnl_matrix_fixed(vnl_matrix_fixed&& other) noexcept = default;
   //: Copy another vnl_matrix_fixed<T,m,n> into this.
   vnl_matrix_fixed& operator=(const vnl_matrix_fixed& rhs) = default;
-  vnl_matrix_fixed& operator=(vnl_matrix_fixed&& rhs) noexcept = default;
+
+  // NOTE: move-assignment throw an exception, because we need to maintain
```
dzenanz: throw -> throw**s**

In other places too

core/vnl/algo/vnl_matrix_inverse.h
```diff
@@ -34,7 +34,14 @@ struct vnl_matrix_inverse : public vnl_svd<T>
   vnl_matrix_inverse(vnl_matrix<T> const & M): vnl_svd<T>(M) { }
   ~vnl_matrix_inverse() override = default;
 
-  operator vnl_matrix<T> () const { return this->inverse(); }
+  vnl_matrix<T> as_matrix() const { return this->inverse(); }
+
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator vnl_matrix<T>() const { return this->inverse(); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: Why is this commented out?

core/vnl/vnl_diag_matrix.h
```diff
@@ -148,13 +148,23 @@ class VNL_EXPORT vnl_diag_matrix
     return diagonal_[r];
   }
 
-  // Need this until we add a vnl_diag_matrix ctor to vnl_matrix;
-  inline vnl_matrix<T> asMatrix() const;
+  //VXL_DEPRECATED_MSG("Deprecated inconsistent return type.\nWARNING: .as_ref returns a vnl_matrix, not a vnl_matrix_ref.")
```
dzenanz: un-comment?

core/vnl/vnl_diag_matrix.h
```diff
@@ -148,13 +148,23 @@ class VNL_EXPORT vnl_diag_matrix
...
+  VXL_DEPRECATED_MSG("Deprecated inconsistent name.\nUSE: .as_matrix() new consistent name.")
+  vnl_matrix<T> asMatrix () const { return this->as_matrix(); }
+#endif
 
   // This is as good as a vnl_diag_matrix ctor for vnl_matrix:
-  inline operator vnl_matrix<T> () const { return asMatrix(); }
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator vnl_matrix<T> () const { return this->as_matrix(); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_matrix_fixed.h
```diff
@@ -660,28 +660,27 @@ class VNL_EXPORT vnl_matrix_fixed
...
 
   //: Cheap conversion to vnl_matrix_ref
   // Sometimes, such as with templated functions, the compiler cannot
   // use this user-defined conversion. For those cases, use the
   // explicit as_ref() method instead.
-  inline operator const vnl_matrix_ref<T>() const { return vnl_matrix_ref<T>( num_rows, num_cols, const_cast<T*>(data_block()) ); }
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator const vnl_matrix_ref<T>() const { return  this->as_ref(); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_matrix_fixed_ref.h
```diff
@@ -708,28 +708,26 @@ class VNL_EXPORT vnl_matrix_fixed_ref : public vnl_matrix_fixed_ref_const<T,num_
...
+  vnl_matrix<T> as_matrix() const { return vnl_matrix<T>(const_cast<T*>(data_block()),num_rows,num_cols); }
 
   //: Cheap conversion to vnl_matrix_ref
   // Sometimes, such as with templated functions, the compiler cannot
   // use this user-defined conversion. For those cases, use the
   // explicit as_ref() method instead.
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator const vnl_matrix_ref<T>() const { return vnl_matrix_ref<T>( num_rows, num_cols, const_cast<T*>(data_block()) ); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_transpose.h
```diff
@@ -44,13 +44,24 @@ class VNL_EXPORT vnl_transpose
   vnl_transpose(const vnl_matrix<double>& M): M_(M) {}
 
   //: Noisily convert a vnl_transpose to a matrix
+
+
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator vnl_matrix<double> () const { return M_.transpose(); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_vector_fixed.h
```diff
@@ -260,26 +260,26 @@ class VNL_EXPORT vnl_vector_fixed
...
   //: Cheap conversion to vnl_vector_ref
   // Sometimes, such as with templated functions, the compiler cannot
   // use this user-defined conversion. For those cases, use the
   // explicit as_ref() method instead.
-  operator const vnl_vector_ref<T>() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
-
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator const vnl_vector_ref<T>() const { return this->as_ref(); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_vector_fixed_ref.h
```diff
@@ -85,18 +85,27 @@ class VNL_EXPORT vnl_vector_fixed_ref_const
...
   //: Cheap conversion to vnl_vector_ref
   // Sometimes, such as with templated functions, the compiler cannot
   // use this user-defined conversion. For those cases, use the
   // explicit as_ref() method instead.
-  operator const vnl_vector_ref<T>() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
+  explicit operator vnl_vector_ref<T>() { return this->as_ref(); }
+#if VXL_LEGACY_FUTURE_REMOVE
+  explicit operator const vnl_vector_ref<T>() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
+#else
+  //VXL_DEPRECATED_MSG("Implicit cast conversion is dangerous.\nUSE: .as_vector() or .as_ref() member function for clarity.")
```
dzenanz: un-comment?

core/vnl/vnl_matrix_fixed.h
hjmjohnson: Fixed.  An improved grammar of sentence to be more expressive.

core/vnl/algo/vnl_matrix_inverse.h
hjmjohnson: Fixed and improved upon to allow selection of how conversions should occur.  Needed to add a separate option for 

```
# Historically VXL implicitly converts between vnl_vector_fixed and vnl_vector or vnl_vector_ref ( & matrix types too),
# when VXL_USE_HISTORICAL_IMPLICIT_CONVERSIONS:BOOL=OFF, require explicit conversions to occur, this simplifies finding
# performance eating conversions by having the compiler throw errors until an explicit conversion is put into place.
option(VXL_USE_HISTORICAL_IMPLICIT_CONVERSIONS "Set backward compatible implicit conversion of types without warning or error when ON" ON)
mark_as_advanced(VXL_USE_HISTORICAL_IMPLICIT_CONVERSIONS)
```

hjmjohnson:
@dzenanz I think I have addressed all your fabulous critiques.  

The only non-comment changes are the last two patches to update the VXL version number and to add fine-grained control of future features.  We need these new options to maintain reasonable backward compatibility (as indicated by ITK building).

core/vnl/vnl_vector.h
hjmjohnson: Fixed, but forgot missed adding the patch.  Will add now.

core/vnl/vnl_vector.h
hjmjohnson: Done.

core/vnl/vnl_matrix.h
hjmjohnson: done.

PR #696, opened 2019-10-02
ENH: travis CI update (coverage & cache)

drewgilliam:
Travis CI update including:
- Fix incremental build issue 
  - cached build directory `$PROJ_BUILD_DIR` was @ `/home/travis/build`
  - travis stores the source repository in `$TRAVIS_BUILD_DIR`, which turns out to be @ `/home/travis/build/<USER>/<REPO>`
  - source repo was thus a subdir of cache build directory, and cache would overwrite source repo on restore
  - move build cache `$PROJ_BUILD_DIR` to non-conflicting location `/home/travis/proj/build`
- Ignore cache on pull request or cron job (i.e., always perform a full rebuild before accepting a PR or for a regularly scheduled build)
- Expand travis coverage to test other gcc versions (4, 6, 7) based on
https://docs.travis-ci.com/user/languages/c/#gcc-on-linux

Note that expanded gcc coverage revealed a missing header file in vul_string.h (stdexcept) that caused an issue only in gcc 4.9.

Should incremental builds prove successful now that the caching issue is solved, full rebuild on PR can be reconsidered in the future.  For now, users can still take advantage of caching to speed travis tests during development. Users with their own forks are free to clear their cache at will as per
https://docs.travis-ci.com/user/caching/#clearing-caches

## PR Checklist
🚫 Makes breaking changes to the vxl/core/* API that requires semantic versioning increase
🚫 Makes design changes to existing vxl/core* API that requires semantic versioning increase
✔️ Adds tests and baseline comparison (quantitative).
🚫 Adds Documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,133 @@
 #Travis CI configuration for VXL
 # See http://travis-ci.org/vxl/vxl
 
-#dist: trusty # This is the default as of August 2017
+#dist: xenial # This is the default as of April 2019
 sudo: false
-
 language: cpp
-compiler:
-  - clang
-  - gcc
 
-# environment variables
+# global environment variables
+# Note source code is checked out by travis to $TRAVIS_BUILD_DIR,
+# located at ${HOME}/build/<USER>/<REPO>.
+# $PROJ_BUILD_DIR must not be co-located with the source directory!
 env:
   global:
-    - PROJ_SOURCE_DIR=${TRAVIS_BUILD_DIR}
-    - PROJ_BUILD_DIR=${HOME}/build
-    - PROJ_INSTALL_DIR=${HOME}/install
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@drewgilliam FYI:  The caching does not seem to be helping the run times.

Making non-code changes seems to result in rebuilds that take the same amount of time.

 18 min 57 sec

drewgilliam:
If you mean the caching is not decreasing runtime for this PR every time you merge in master, that's expected.  This is TRAVIS_BUILD_TYPE == pull_request, and the system will currently perform a full build for every PR.   Only on regular pushes (like if you're developing a branch in this repo or a fork) would the cache be used.

hjmjohnson:
> If you mean the caching is not decreasing runtime for this PR every time you merge in master, that's expected. This is TRAVIS_BUILD_TYPE == pull_request, and the system will currently perform a full build for every PR. Only on regular pushes (like if you're developing a branch in this repo or a fork) would the cache be used.

Thanks for the explanation.

PR #699, opened 2019-10-04
Minor preparatory changes

hjmjohnson:
Smaller changes identified during review and preparation for adding move semantics to vnl_vector and vnl_matrix.

### DIFF
--- a/core/vnl/vnl_matrix_fixed.h
+++ b/core/vnl/vnl_matrix_fixed.h
@@ -393,6 +393,11 @@ class VNL_EXPORT vnl_matrix_fixed
   //: Set values of this matrix to those of M, starting at [top,left]
   vnl_matrix_fixed& update(vnl_matrix<T> const&, unsigned top=0, unsigned left=0);
 
+  //: Set values of this matrix to those of M, starting at [top,left]
+  vnl_matrix_fixed<T,num_rows,num_cols>& update(
+      vnl_matrix_fixed<T,num_rows,num_cols> const& m,
+      unsigned top=0, unsigned left=0);
+
   //: Set the elements of the i'th column to v[i]  (No bounds checking)
   vnl_matrix_fixed& set_column(unsigned i, T const * v);
 
--- a/core/vnl/vnl_matrix_fixed.h
+++ b/core/vnl/vnl_matrix_fixed.h
@@ -440,10 +440,10 @@ class VNL_EXPORT vnl_matrix_fixed
   vnl_vector_fixed<T,num_rows> get_column(unsigned col) const;
 
   //: Get a matrix composed of rows from the indices specified in the supplied vector.
-  vnl_matrix<T> get_rows(vnl_vector<unsigned int> i) const;
+  vnl_matrix<T> get_rows(const vnl_vector<unsigned int> &i) const;
 
   //: Get a matrix composed of columns from the indices specified in the supplied vector.
-  vnl_matrix<T> get_columns(vnl_vector<unsigned int> i) const;
+  vnl_matrix<T> get_columns(const vnl_vector<unsigned int> &i) const;
 
   //: Get n rows beginning at rowstart
   vnl_matrix<T> get_n_rows   (unsigned rowstart, unsigned n) const;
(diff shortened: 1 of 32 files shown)


hjmjohnson:
@N-Dekker Would you mind reviewing this PR?  It should be low hanging fruit and not too controversial.

core/vnl/vnl_vector.hxx
```diff
@@ -96,10 +96,9 @@ template<class T>
 vnl_vector<T>::vnl_vector (size_t len, size_t n, T const values[])
 {
   vnl_vector_alloc_blah(len);
-  if (n > 0) {                                  // If user specified values
-    for (size_t i = 0; i < len && n; i++, n--)        // Initialize first n elements
-      this->data[i] = values[i];                // with values
-  }
+  // If user specified values, initialize first n elements with values
+  const size_t copy_num{std::min(len,n) };
+  std::copy(values, values + copy_num, data);
```
N-Dekker: Looks OK, but C++11 `std::copy_n` seems even better to me: 

    std::copy_n(values, copy_num, data);

core/vnl/vnl_matrix_fixed.h
```diff
@@ -667,16 +667,16 @@ class VNL_EXPORT vnl_matrix_fixed
   // for vnl_matrix but not for vnl_matrix_fixed. There is also a
   // conversion operator that should work most of the time.
   // \sa vnl_matrix_ref::non_const
-  inline const vnl_matrix_ref<T> as_ref() const { return vnl_matrix_ref<T>( num_rows, num_cols, const_cast<T*>(data_block()) ); }
```
N-Dekker: I'm afraid this unusual 'const'-qualification at the top level of the return type does have a meaning here! You see, this is the `const` overload of `as_ref()`. The top-level `const` should protect against modifying the original matrix through the `vnl_matrix_ref`.

core/vnl/vnl_vector_fixed_ref.h
```diff
@@ -90,13 +90,13 @@ class VNL_EXPORT vnl_vector_fixed_ref_const
...
   // conversion operator that should work most of the time.
   // \sa vnl_vector_ref::non_const
-  const vnl_vector_ref<T> as_ref() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
+  vnl_vector_ref<T> as_ref() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
 
   //: Cheap conversion to vnl_vector_ref
   // Sometimes, such as with templated functions, the compiler cannot
   // use this user-defined conversion. For those cases, use the
   // explicit as_ref() method instead.
-  operator const vnl_vector_ref<T>() const { return vnl_vector_ref<T>( n, const_cast<T*>(data_) ); }
```
N-Dekker: As I commented before: the unusual 'const'-qualification at the top level of the return type appears to be meant to prevent modifications of the original object (vnl_vector_fixed, in this case) through the "ref" object.

core/vnl/vnl_vector.hxx
hjmjohnson: I put that on a todo.  There are MANY places that this improvement could be changed, and I'd rather do that uniformly.

core/vnl/vnl_matrix_fixed.h
hjmjohnson: Reverted:  FYI That is part of another larger issue that I want to keep separate.

the larger issue is my concern about excessive unnecessary implicit type conversions.

core/vnl/vnl_vector_fixed_ref.h
hjmjohnson: Reverted:  FYI That is part of another larger issue that I want to keep separate.

the larger issue is my concern about excessive unnecessary implicit type conversions.

core/vnl/vnl_matrix_fixed.hxx
```diff
@@ -516,7 +516,7 @@ vnl_matrix_fixed<T,nrows,ncols>
 template <class T, unsigned int nrows, unsigned int ncols>
 vnl_matrix<T>
 vnl_matrix_fixed<T,nrows,ncols>
-::get_columns(vnl_vector<unsigned int> i) const
+::get_columns(const vnl_vector<unsigned int> & i) const
```
N-Dekker: Good catch! 👍

hjmjohnson: Thanks.

PR #700, opened 2019-10-09
BUG: vnl_matrix & vnl_vector swap

drewgilliam:
PR #694 recently added a `m_LetArrayManageMemory` member variable to `vnl_matrix` and `vnl_vector`.  Add this new member variable to respective `swap` functions.

## PR Checklist
🚫 Makes design changes to existing vxl/core/\* API that requires semantic versioning increase
🚫 Adds tests and baseline comparison (quantitative).
🚫 Adds Documentation.


drewgilliam:
@hjmjohnson

PR #701, opened 2019-10-09
BUG: vnl_matrix move-assignment segfault

drewgilliam:
In the move-assignment `operator=` for `vnl_matrix` (see PR #694) when both left-hand side and right-hand side objects manage their own memory, the right-hand side object should finally be invalidated with null data, zero size, and  `m_LetArrayManageMemory` set to a default value of true (not false).  The system segfaults should a user attempt to `std::swap` two matrices.

Set `m_LetArrayManageMemory` to true in move-assignment `operator=` for both `vnl_matrix` and `vnl_vector`.  Add more swap tests, including native swap and `std::swap` for both `vnl_matrix` and `vnl_vector`.

## More Details
During move-assignment, `vnl_matrix` wrongly reset `m_LetArrayManageMemory` to false.  This is an issue should a user attempt to `std:swap` two matrices (while users should prefer the native `vnl_matrix::swap` function, `std::swap` should certainly not segfault).  

The C++11 `std::swap` method is implemented as follows
```
template<typename T> void swap(T& t1, T& t2) {
    T temp = std::move(t1); // line (1)
    t1 = std::move(t2);     // line (2)
    t2 = std::move(temp);   // line (3)
}
```

When using `std::swap` on a `vnl_matrix`, matrix t1 would be cleared after line 1 but with `m_LetArrayManageMemory = false`.  Line two would attempt to follow a code path for externally managed memory and the system would segfault.

## PR Checklist
🚫 Makes design changes to existing vxl/core\* API that requires semantic versioning increase
✔️ Adds tests and baseline comparison (quantitative).
🚫 Adds Documentation.

### DIFF
--- a/core/vnl/tests/test_container_interface.cxx
+++ b/core/vnl/tests/test_container_interface.cxx
@@ -121,15 +121,21 @@ void test_common_interface()
     TContainer l(2, 2);
     l.copy_in(data1);
     TContainer l_swap(l);
+    TContainer l_std_swap(l);
 
     const typename TContainer::element_type data2[4] = {4, 5, 6, 7};
     TContainer r(2, 2);
     r.copy_in(data2);
     TContainer r_swap(r);
+    TContainer r_std_swap(r);
 
     l_swap.swap(r_swap);
     TEST("swap left-right", l.is_equal(r_swap, 10e-6), true);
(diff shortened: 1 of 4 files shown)


drewgilliam:
@hjmjohnson

PR #703, opened 2019-10-11
Fix failing build

chris1248:
* COMP: Many compiler errors like this:

error C2039: 'back_inserter': is not a member of 'std'

Solution:
Add the <iterator> header to solve.


Altair3:
+1

PR #711, opened 2019-10-31
BUG: vcl: install compiler headers under the include directory

mathstuf:
They show up in the `${prefix}/vcl` right now.

## PR Checklist


mathstuf:
Looks like the GCC 7 CI builder has a bad configuration (the package isn't available anymore?).

PR #720, opened 2019-11-18
DOC: Recommend explicitly using vnl_matrix_inverse.as_matrix()

blowekamp:
The as_vector method does not exist.

## PR Checklist

-  Adds Documentation.


blowekamp:
@hjmjohnson

PR #721, opened 2019-11-19
Fix trailing whitespace

hjmjohnson:
## PR Checklist
-  :no_entry_sign:  Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
- :no_entry_sign:  Makes design changes to existing vxl/core\* API that requires semantic versioning increase


hjmjohnson:
@blowekamp FYI: A very quick patch set to remove trailing whitespace from files.

PR #726, opened 2019-11-24
BUG: Fix memory leak bug & fix compilation dependancies

hjmjohnson:
@blowekamp @dzenanz @thewtex 

The memory leak fix is relatively small, but testing it turned out to be very confusing because of the changes to the vnl_matrix.hxx file did not include a recompilation of the explicit instantiation or the tests.

This confounding build bug was due to the fact that the system includes with `#include <>` do not generate dependencies in the build tree like local includes with `#include ""` do.


core/vnl/vnl_matrix.hxx
```diff
@@ -265,10 +265,8 @@ vnl_matrix<T>& vnl_matrix<T>::operator=(vnl_matrix<T>&& rhs)
     else
     {
       // Release any resource we're previously holding
-      if(this->data)
-      {
-        vnl_c_vector<T>::deallocate(this->data[0], this->num_rows*this->num_cols);
-      }
+      vnl_matrix_free_blah;
```
hjmjohnson: @blowekamp, @dzenanz  This is the key change that fixes the bug reported in https://github.com/InsightSoftwareConsortium/ITK/issues/1438

The other changes were necessary to enable tracking this down because the builds were not refecting this code change due to lack of rebuild when this file was changed.

dzenanz:
Good work Hans! I hoped it would fix https://github.com/InsightSoftwareConsortium/ITK/issues/1431, but unfortunately it doesn't.

PR #729, opened 2019-12-06
BUG: fixed signed integer overflow by casting to unsigned long

seanm:
UBSan found a signed integer overflow. Arithmetic with 6 unsigned shorts gets promoted to signed int, which can then overflow with big numbers. Cast to unsigned long, as in lines preceding, to force unsigned arithmetic, where overflow is well-defined.

### DIFF
--- a/core/vnl/vnl_drand48.cpp
+++ b/core/vnl/vnl_drand48.cpp
@@ -153,7 +153,9 @@ _dorand48(unsigned short xseed[3])
 					(unsigned long) _rand48_mult[1] * (unsigned long) xseed[0];
 	temp[1] = (unsigned short) accu;        /* middle 16 bits */
 	accu >>= sizeof(unsigned short) * 8;
-	accu += _rand48_mult[0] * xseed[2] + _rand48_mult[1] * xseed[1] + _rand48_mult[2] * xseed[0];
+	accu += (unsigned long)_rand48_mult[0] * (unsigned long)xseed[2] +
+					(unsigned long)_rand48_mult[1] * (unsigned long)xseed[1] +
+					(unsigned long)_rand48_mult[2] * (unsigned long)xseed[0];
 	xseed[0] = temp[0];
 	xseed[1] = temp[1];
 	xseed[2] = (unsigned short) accu;


seanm:
@N-Dekker @vfonov @thewtex @dzenanz this fixes a couple of ITK test failures, ex:

https://open.cdash.org/testDetails.php?test=829115206&build=6239398

core/vnl/vnl_drand48.cpp
```diff
@@ -153,7 +153,9 @@ _dorand48(unsigned short xseed[3])
 					(unsigned long) _rand48_mult[1] * (unsigned long) xseed[0];
 	temp[1] = (unsigned short) accu;        /* middle 16 bits */
 	accu >>= sizeof(unsigned short) * 8;
-	accu += _rand48_mult[0] * xseed[2] + _rand48_mult[1] * xseed[1] + _rand48_mult[2] * xseed[0];
+	accu += (unsigned long)_rand48_mult[0] * (unsigned long)xseed[2] +
+					(unsigned long)_rand48_mult[1] * (unsigned long)xseed[1] +
+					(unsigned long)_rand48_mult[2] * (unsigned long)xseed[0];
```
N-Dekker: I'm sorry @seanm, but seeing so many C-style casts (unsigned long) makes me very unhappy. Even though the old code already has them. It's now 2019, and we use C++11! The safe C++11 way to do type cast is by _curly braces_: `T{value}`. If that's not possible, you may consider `static_cast<T>(value)`. 

Another issue I see here is that `unsigned long` is different of different platforms. On Windows/Visual Studio it is 32-bit, while on Linux/GCC it is 64-bit. It's unclear to me now what is the intended number of bits here.

N-Dekker: It looks like it should be OK to use `std::uint32_t`, instead of `unsigned long`, in this function (`_dorand48`). But if you want to be really sure you don't get overflow, use `std::uint64_t`.

Then of course, it would be more clear to use `std::uint16_t`, instead of `unsigned short`. 

Conversions from 16-bit unsigned to 32 or 64 can be done using C++11 curly braced conversion `T{value}`. From 32 or 64 to 16-bit could use `static_cast<std::uint16_t>(value)`.

seanm: The number of bits isn't so relevant, the most important thing is that the cast makes things unsigned (and at least as large as int). As you see a few lines above, those same values (`_rand48_mult` and `xseed`) are already cast to unsigned long and `accu` itself is also unsigned long.

As for the casting style, I think it makes more sense to keep with the existing code style of that function and file.

hjmjohnson: I agree with both of you for different reasons.

I will merge this as is (@seanm) as it makes the minimal changes necessary to fix the stated problem.

I will make a new PR to address the "modernization" concerns identified by @N-Dekker .

N-Dekker: I tried the following on Visual C++ 2017, using Code Analysis:

    int main(int argc, char**)
    {
        unsigned u = (unsigned)argc;
    }

It says:
> warning C26493: Don't use C-style casts (type.4).

seanm: You realize this is vxl, not ITK, right? :)

You could spend many many many manhours modernizing it.

Isn't ITK moving away from vxl anyway?

hjmjohnson: YES!  Hopefully anyway :).  I have spent many/many/many hours modernizing it.  

NOTE:  This is a copy of the posix version of that same function needed to satisfy non-posix machines.  Modernizing also means diverging from the original upstream source code.

seanm: upstream where? Isn't this vxl's home? Not sure I understand...

N-Dekker: Actually I saw already some modern C++11 `constexpr` in there:

  https://github.com/vxl/vxl/blob/master/core/vnl/vnl_drand48.cpp#L21

The only valid reason for me to add any extra C-style cast in my code would be for C compatibility.  But clearly, C compatibility was abandoned already 😃  More C-style casts will eventually just produce more warnings, instead of less.

seanm:
So some additional magic is needed to get this into ITK, right?

Issue #734, opened 2019-12-10
Problem with strings.h on VS2019


mundyj:
### Expected behavior

geotiff should compile on visual studio
### Actual behavior
D:\software\terra_dsm\external\vxl\v3p\geotiff\cpl_serv.h(45,12): fatal error C1083: Cannot open include file: 'strings.h': No such file or directory

```none
```

### Reproducibility

100% since strings.h is not available
### Versions
After pull request 725

### Environment
Windows 10, VS2019 16.3.9

hjmjohnson:
Closing as obsolete. The strings.h include path that triggered this on VS2019 in 2019 is no longer in current source; CI on Windows 2022 is clean.

PR #735, opened 2019-12-10
Revert "COMP: Some compilers need both string.h & strings.h"

hjmjohnson:
This reverts commit 901b41bb412ec18e12a7801a331a19e877b55ba3.

This resolves issue #734 of compilation failure, but
re-introduces a compiler warning about implicitly defined
function on some compilers.

### DIFF
--- a/v3p/geotiff/cpl_serv.h
+++ b/v3p/geotiff/cpl_serv.h
@@ -41,9 +41,10 @@
 #ifdef HAVE_STRING_H
 #  include <string.h>
 #endif
-#if defined(HAVE_STRINGS_H)
+#if defined(HAVE_STRINGS_H) && !defined(HAVE_STRING_H)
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@mundyj  could you please confirm that this new code resolves your problem on windows?

mundyj:
That did fix the problem but this branch caused the following error
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2039	'min': is not a member of 'std' (compiling source file D:\software\terra_dsm\external\vxl\core\vnl\vnl_bignum_traits.cxx)	vnl	D:\software\terra_dsm\external\vxl\core\vnl\vnl_bignum.h	222	

If I include <algorithm> in vnl_bignum.h the problem is resolved.
Again, this is with Visual Studio 2019
Thanks!

PR #737, opened 2019-12-10
BUG: Missing header for std::min on visual studio.

hjmjohnson:


hjmjohnson:
@mundyj Does this fix the next problem?

mundyj:
Hans,
Yes, visual-studio-header-missing solves the problem.
Thanks,
Joe

PR #738, opened 2019-12-11
Fix test timing

hjmjohnson:


hjmjohnson:
@thewtex This might fix the error.   I'm concerned that this type of tests may fail if the machine is under heavy load.

Previously the test times were saturated with the cost of 6 random number generations, assignments, in addition to the running of the function to be tested.

Hans

PR #774, opened 2020-04-14
Line endings fix

dzenanz:
## PR Checklist

-  Adds Documentation.


dzenanz:
@hjmjohnson consider merging and integrating into your ITK PR https://github.com/InsightSoftwareConsortium/ITK/pull/1767

PR #781, opened 2020-05-20
BUG: acal match_tree destructor

NoahRJohnson:
## PR Checklist

-  :no_entry_sign: Makes changes to the contributed directory API DOES NOT require semantic versioning increase

Adds a guard to the acal_match_tree destructor to prevent segfaults when root_ hasn't been initialized.


NoahRJohnson:
Instead of adding a bunch of guards, I've removed the default constructor and required a root node to construct a tree.

PR #788, opened 2020-06-27
added multiple ray intersection with covariance

mundyj:
## PR Checklist

- :no_entry_sign: --> Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
- [X] : --> Makes design changes to existing vxl/core\* API that requires semantic versioning increase

-   :no_entry_sign: or  Makes changes to the contributed directory API DOES NOT require semantic versioning increase
-  [X] or  Adds tests and baseline comparison (quantitative).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ foreach(p
 endforeach()
 
 project(VXL #Project name must be all caps to have properly generated VXL_VERSION_* variables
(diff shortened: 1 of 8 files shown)


decrispell:
I believe this PR should include a bump in the semantic version number because it adds a new vgl_intersection function in core (design change in existing vxl/core* API).

PR #796, opened 2020-09-24
COMP: Remove implicit addition of -msse2 flag

hjmjohnson:
If sse2 build options are requested, require that they
be requested additions of compiler flags.


hjmjohnson:
@dzenanz I don't think this fixes the ITK problem, but it does make building cleaning by not magically adding -msse2 flags with a backdoor add_definition from 2007, 2008.

dzenanz (approved):

PR #799, opened 2020-09-29
ENH: append pointsets

scottsorensenvision:
Adding a warning message to appending pointsets with and without normals. There is not really a good solution to what to do in this case, but it should not silently fail to append.

## PR Checklist

- :no_entry_sign:Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
- :no_entry_sign:  Makes changes to the contributed directory API DOES NOT require semantic versioning increase


decrispell (approved):

decrispell:
test failure looks unrelated.

PR #807, opened 2020-10-22
BUG: contrib/brl std::is_same

drewgilliam:
bsgm_prob_pairwise_dsm bug fix - favor `std::is_same` over other methods for template type discovery.  Previous method attempting to compare `std::numeric_limits<PIX_T>::max()` to expected maximum value does not produce the expected result for `PIX_T==unsigned char`.
https://en.cppreference.com/w/cpp/types/is_same

@decrispell 

## PR Checklist
- ❌ Makes design changes to existing vxl/core\* API that requires semantic versioning increase
- ❌  Makes changes to the contributed directory API DOES NOT require semantic versioning increase
- ❌  Adds tests and baseline comparison (quantitative).
- ❌  Adds Documentation.


drewgilliam:
FYI @mundyj

PR #810, opened 2020-11-19
ENH: Add front() and back() members to vnl_vector and vnl_vector_fixed

hjmjohnson:
Replace `vnl_vector[vnl_vector.size() - 1]` by vnl_vector.back().
Also add vnl_vector::front()"

Co-authored-by: Niels Dekker <N.Dekker@lumc.nl>


hjmjohnson:
@N-Dekker  Here is the promised upstream patch!

PR #818, opened 2020-12-16
COMP: Remove unused netlib function

hjmjohnson:
The adaquad.c file was only used in a test inside
of the netlib v3p folder.  The algorithm was not used
in any other part of vxl.

The adaquad function was identified as having a memory
array bounds overrun bug (from the netlib test) when
included in ITK.

Removing this interface has no impact on VXL, or any
of it's included contrib libraries.


hjmjohnson:
@seanm  Removing one of the Common failures on ITK.  

https://open.cdash.org/test/305444698

A useless bit of code is failing.  It appears that this code is only used to run the test that fails.  It has no other use.

hjmjohnson:
Note: I double-checked, and this is also not used in ITK.

seanm:
I really don't know the vnl codebase, but seems like a reasonable solution.

Thanks for looking into it!  Will be nice to one day have no ASan failures.

PR #823, opened 2021-02-23
COMP: Address Wshadow Compiler Warnings

DVigneault:
The local typedefs "abs_t" and "real_t" both shadow
typedefs declared in the classe's header file.  This
patch removes one redundant typedef, and renames another
so as not to conflict with the same-named typedef
in the parent class.  These warnings were observed on
Arch Linux using g++ version 10.2.0.

## PR Checklist

-  Adds Documentation.


DVigneault:
@hjmjohnson @thewtex Just addressing a few warnings I encountered while building ITK.

thewtex (approved):
@DVigneault thanks!

@hjmjohnson please take a look.

hjmjohnson (approved):
LGTM

Issue #828, opened 2021-03-12
The value of sqrt(2π) is wrong


Leengit:
### Description
The value of sqrt(2π) is wrong.

### Expected behavior
core/vnl/vnl_math.h:113 should be
```c++
static constexpr double sqrt2pi          = 2.50662827463100050242;
```
See http://oeis.org/A019727/constant or compute it yourself with a high-precision calculator.

This value converted to double precision is 5,644,425,081,792,262×2^(-51) and is about 1.83e-16 larger than the exact value of sqrt(2π), which is approximately 5,644,425,081,792,261.587×2^(-51).

### Actual behavior
```c++
static constexpr double sqrt2pi          = 2.50662827463100024161;
```

This value converted to double precision is 5,644,425,081,792,261×2^(-51) and is  2.61e-16 smaller than the exact value of sqrt(2π).  This difference is worse than the difference for the expected behavior.

### Reproducibility

100%

### Versions
Current master branch March 11, 2021.

dzenanz:
How did you discover this?

Leengit:
I discovered that R would give me these mathematical values to arbitrary precision so I thought to try 50 digits of precision and zero tolerance for error in the `TEST_CONSTANT` macro of the ITK version of the `test_math.cxx` file.  Despite that some of the constants were not correctly represented as decimals in `vnl_math.h`, only `sqrt2pi` failed the newly stringent test.  (The other "wrong" decimal values nonetheless gave the same double-precision value as the correct decimal values.)

dzenanz:
I think this can now be closed.

PR #829, opened 2021-03-12
BUG: Fix the value of sqrt2pi

Leengit:
Resolves Issue #828.  Changes
```c++
static constexpr double sqrt2pi          = 2.50662827463100024161;
```
to
```c++
static constexpr double sqrt2pi          = 2.50662827463100050242;
```
Is the change of value considered a breaking change?


Leengit:
With this second commit, I changed the new value of `sqrt2pi` to its rounded value rather than its truncated value.  I also extended some constants by a digit or two so that all constants have the same number of significant digits.  In neither case is the double-precision value affected, so this second commit is purely about STYLE.

On the other hand, the first commit does change the least-significant bit of the double-precision representation of `sqrt2pi`.  That is not merely cosmetic.

hjmjohnson (commented):
Please add a comment in the code describing how the correct values were identified.  What source did you use?  Are these base 10 values the closest correct representation in base 2 IEEE 754 format?

Leengit:
The second commit is cosmetic.  It changes some insignificant decimal digits of the constants to be their correct, rounded values to that precision.  However, the defined double-precision and single-precision constants are unchanged by these cosmetic changes.

Leengit:
> Please add a comment in the code describing how the correct values were identified. What source did you use? Are these base 10 values the closest correct representation in base 2 IEEE 754 format?

I will also add something to the code.

I computed them with R's Rmpfr package and checked the results against oeis.org.  In every case, the base 10 value is the true value, rounded to the number of decimal digits supplied.  In every case except `sqrt2pi` I verified that the resulting double-precision value is unchanged.  For `sqrt2pi` I verified that the new value's double-precision representation is closer to the true value than the previous value.

You could try the following code with R (for instance [here](https://rdrr.io/snippets/)):
```
library(Rmpfr)
bits <- 1000
one <- mpfr(1, bits)
Pi <- 4*atan(one)
e <- exp(one)
log2e <- 1 / log(2*one)
log10e <- 1 / log(10*one)
ln2 <- log(2*one)
ln10 <- log(10*one)
twopi = 2 * Pi
pi_over_2 = Pi / 2
pi_over_4 = Pi / 4
pi_over_180 = Pi / 180
one_over_pi = 1 / Pi
two_over_pi = 2 / Pi
deg_per_rad = 180 / Pi
sqrt2pi = sqrt(2*Pi)
two_over_sqrtpi = 2 / sqrt(Pi)
one_over_sqrt2pi = 1 / sqrt(2 * Pi)
sqrt2 = sqrt(2 * one)
sqrt1_2 = 1 / sqrt(2 * one)
sqrt1_3 = 1 / sqrt(3 * one)
for (var in setdiff(ls(), c("bits", "one"))) {
    cat(sprintf("%16s = %s\n", var, format(get(var), digits=21, scientific=FALSE)))
}
```

hjmjohnson:
@Leengit Thanks!

Leengit:
In the code I just pushed, I have marked each constant with the URL of the sequence at _The On-line Encyclopedia of Integer Sequences_ that confirms the accuracy of those digits.  The values in the code have been rounded to the displayed precision rather than simply truncating the OEIS sequence.

If you would prefer truncation, which works better with string searches, I can change that. 

If you would like the R-code from this pull request to be in the code too, I can do that.

hjmjohnson (approved):

hjmjohnson (changes requested):
This is also going to require update to the version number.  Since these values changed, it will change many external results for other applications that have depended on these values for many years.

I am thinking of ITK in particular.  @dzenanz @thewtex  Please review this PR carefully.  How should this bug, be fixed without disrupting backward compatibility?

Leengit:
I will build ITK from sources with these changes to check for compatability.

dzenanz (approved):
I don't think that change in such a far decimal place is going to cause many regression baseline changes. We don't have many double-precision baseline images, maybe even none.

Leengit:
https://github.com/InsightSoftwareConsortium/ITK/pull/2408 is the pull request that moves the changes from this pull request into ITK for testing purposes.  It is all green at that pull request itself and at [CDash](https://open.cdash.org/index.php?project=Insight&date=2021-03-15&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=2408).

dzenanz:
I think this can be merged.

hjmjohnson (approved):

PR #834, opened 2021-04-05
ENH: vil_flatten & vil_math_value_range_percentiles(ignore_nan)

drewgilliam:
ENH: add `vil_flatten` to copy `vil_image_view` data into a 1D `std::vector`.  Provides both row-major and column-major implementations.  Add tests.

ENH: add `ignore_nan` argument to `vil_math_value_range_percentiles`, allowing users to process floating point images containing nans.  Add/expand tests.

## PR Checklist

- ✔️ Makes design changes to existing vxl/core\* API that requires semantic versioning increase
- ❌ Makes changes to the contributed directory API DOES NOT require semantic versioning increase
- ✔️ Adds tests and baseline comparison (quantitative).
- ❌ Adds Documentation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ foreach(p
 endforeach()
 
 project(VXL #Project name must be all caps to have properly generated VXL_VERSION_* variables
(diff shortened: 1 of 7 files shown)


drewgilliam:
@decrispell

decrispell (approved):

PR #841, opened 2021-05-23
map focal plane rectification homograpies to common image space

mundyj:
## PR Checklist

-  :no_entry_sign: --> Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
-  :no_entry_sign: --> Makes design changes to existing vxl/core\* API that requires semantic versioning increase

-  :no_entry_sign: --> Makes changes to the contributed directory API DOES NOT require semantic versioning increase
-  :no_entry_sign: --> Adds tests and baseline comparison (quantitative).

### DIFF
--- a/core/vpgl/algo/vpgl_equi_rectification.cxx
+++ b/core/vpgl/algo/vpgl_equi_rectification.cxx
@@ -334,9 +334,11 @@ bool vpgl_equi_rectification::rectify_pair(vpgl_perspective_camera<double> const
     return false;
   }
   // update the rectificaion homographies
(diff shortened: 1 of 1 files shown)


mundyj:
Found a bug in perspective rectification that occurs if the stereo pair images have different calibration matrices. Normally for video cameras they are always the same and so the problem didn't show up until a recent experiment

decrispell (approved):

mundyj:
Dan,
Let me know when it is merged.
Thanks,
Joe

PR #844, opened 2021-06-09
BUG/STYLE: contrib/brl acal single node tree

drewgilliam:
BUG: fix `acal_match_graph::validate_match_trees_and_set_metric` for trees with only a single node (parent discovery for a single-node tree resulted in a segfault)

STYLE: general formatting in `acal_match_graph` and `acal_match_tree`

## PR Checklist

- ❌ Makes design changes to existing vxl/core\* API that requires semantic versioning increase
- ✔️ Makes changes to the contributed directory API DOES NOT require semantic versioning increase
- ❌ Adds tests and baseline comparison (quantitative).
- ❌ Adds Documentation.


drewgilliam:
@decrispell

decrispell (approved):

PR #850, opened 2021-07-15
BUG: fix a crash due to null pointer dereference

dzenanz:
This would happen in ITK, in
Modules\Numerics\Statistics\include\itkGaussianMixtureModelComponent.hxx
at line m_Covariance.SetIdentity();
and in a few other places too. See:
https://open.cdash.org/viewTest.php?onlyfailed&buildid=7341011
List of tests that Completed (SEGFAULT) copy-pasted here in case of link rot:

itkExpectationMaximizationMixtureModelEstimatorTest
itkGaussianMixtureModelComponentTest
itkImageClassifierFilterTest
itkSampleClassifierFilterTest7

The bug was introduced by c6c899aaf9cbf7ccabfca1ba35b2248fb611ffbc.

### DIFF
--- a/core/vnl/vnl_matrix.hxx
+++ b/core/vnl/vnl_matrix.hxx
@@ -782,10 +782,14 @@ vnl_matrix<T>& vnl_matrix<T>::set_identity()
   // Zero-filling all data, followed by a loop over the diagonal, is
   // probably better than having a branch inside the loop. Similar to
   // the set_identity() member function of vnl_matrix_fixed.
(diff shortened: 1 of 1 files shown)


dzenanz:
@N-Dekker @hjmjohnson

hjmjohnson (approved):

N-Dekker:
Oops, that's my bug 😭 ! Thanks very much for the fix @dzenanz !

PR #857, opened 2021-09-16
TSan issues

seanm:
## PR Checklist

-  Adds Documentation.

### DIFF
--- a/core/vgui/internals/vgui_accelerate_x11.cxx
+++ b/core/vgui/internals/vgui_accelerate_x11.cxx
@@ -8,6 +8,7 @@
 #include <iostream>
 #include <cmath>
 #include <cstring>
+#include <cstdint>
 #include <algorithm>
 #include "vgui_accelerate_x11.h"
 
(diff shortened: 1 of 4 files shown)


seanm:
@hjmjohnson @thewtex @dzenanz

dzenanz (approved):

dzenanz:
```text
/vxl/core/vnl/algo/vnl_svd.hxx:209:39: error: 'atomic_flag_test_and_set' is not a member of 'std'
 inline bool vnl_svn_warned() { return std::atomic_flag_test_and_set(&w); }
```

dzenanz:
Probably needs `#include <atomic>`.

seanm:
> Probably needs `#include <atomic>`.

Doh! Fixed.

dzenanz (approved):

dzenanz:
I don't have write permission here, somebody else will have to merge.

hjmjohnson (approved):

seanm:
There are 3 other tests that fail under ASan:

- vil_test_save_load_image : related to TIFF, but I gave up, the pointer arithmetic is too gnarly
- vil_test_4_plane_tiff : as above
- vpgl_algo_test_rational_adjust : also gnarly code, machine converted from fortran and so hard to decipher

hjmjohnson:
@seanm I don't think those will affect ITK, so I don't think the ITK team will be working on them.  I am not sure which group uses those functions, but perhaps that team could address the issues.

seanm:
The `vpgl_algo_test_rational_adjust` failure I think impacts ITK, because the bug is in `vnl_levenberg_marquardt.cxx` which I believe ITK uses, doesn't it?

seanm:
I created an issue: https://github.com/vxl/vxl/issues/858

PR #859, opened 2021-09-17
BUG: fixed buffer overrun in logging code found by ASan

seanm:
## PR Checklist

-  Adds Documentation.


hjmjohnson (approved):
Spectacular !! Thank you for your hard work in finding and squashing these problems.

PR #864, opened 2021-11-25
Geocamera fix 

mundyj:
## PR Checklist

- :no_entry_sign: --> Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
-  :no_entry_sign: --> Makes design changes to existing vxl/core\* API that requires semantic versioning increase

- :no_entry_sign: --> Makes changes to the contributed directory API DOES NOT require semantic versioning increase
-  :no_entry_sign: --> Adds tests and baseline comparison (quantitative).

### DIFF
--- a/core/vpgl/file_formats/vpgl_geo_camera.cxx
+++ b/core/vpgl/file_formats/vpgl_geo_camera.cxx
@@ -25,7 +25,7 @@
 #include <vil/file_formats/vil_nitf2_image.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


mundyj:
pixel spacing method was returning spacing in degrees rather than meters as specified

mundyj:
Didn't fail due to an error - build timed out due to having a non-paying plan
Comments from the build:
"Upgrade your pricing plan to take advantage of longer build times
context deadline exceeded"

PR #876, opened 2021-12-13
DOC: Change version minor number

hjmjohnson:
VXL minor revision updated to 3.5 in order to allow identification
when vnl_math::abs learned how to consistently process complex numbers,
and the standard types supported by std::abs were preferred over
custom implementations.

## PR Checklist

 [X] Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
 [X] Makes design changes to existing vxl/core\* API that requires semantic versioning increase

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ foreach(p
 endforeach()
 
 project(VXL #Project name must be all caps to have properly generated VXL_VERSION_* variables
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@dzenanz This is needed so that we can get the version of system install VXL to determine how to address https://open.cdash.org/viewBuildError.php?buildid=7621980 in a backward-compatible way.

PR #877, opened 2021-12-13
BUG: Recent assert in debug was too strict

hjmjohnson:
In debug builds, the historical test case would
always fail due to '*n==2' and '*p==2' that did
not pass the 'assert(*n >= *p)' strict condition
added in commit 334e913af16bc9356135da32a45c9980f164cab1.


hjmjohnson:
This issue is showing up in all of the debug builds for ITK https://open.cdash.org/test/563177107#.

@seanm

seanm (approved):

PR #886, opened 2021-12-24
COMP: Remove deprecated WriteCompilerDetectionHeader

hjmjohnson:
The cmake WriteCompilerDetectionHeader mechanism
is deprecated in cmake.

See cmake --help-policy CMP0120


hjmjohnson:
The wrong base directory was used.  This work was already completed.

PR #888, opened 2021-12-24
Limit a test to occur only when long is 64 bit

seanm:
On OpenBSD 7 on PowerPC:

warning: implicit conversion from 'long long' to 'long' changes value from 1234321234321 to 1665620369 [-Wconstant-conversion]

From the surrounding code and comments, it seems to me this code is only supposed to be used when long is 64 bit.  It's 32 bit on OpenBSD PowerPC.


seanm:
@hjmjohnson

hjmjohnson (approved):

PR #895, opened 2022-03-18
COMP: Missing header for std::numeric_limits<>

hjmjohnson:
Clang 13 has fewer implied header includes.

The tests must explicitly include <limits> for the
numeric_limits definitions to avoid pre-processor errors like:

error: no member named 'numeric_limits' in namespace 'std'


hjmjohnson:

dzenanz (approved):

PR #896, opened 2022-03-18
COMP: Avoid implicit conversion value change warning

hjmjohnson:
Clang 13 identifies a value change during an implicit conversion, so
we must set the upperbound carefully.

vxl/v3p/netlib/sparse/spFactor.c:2147:13: warning:
implicit conversion from 'long' to 'double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]


hjmjohnson:

dzenanz (approved):

PR #897, opened 2022-03-18
STYLE: Prefer constexpr for compiletime constants

hjmjohnson:
The initialization values for the the arrays
are hard-coded in the test at compiletime, so
make them constexpr to assist with exposing
memory errors.


hjmjohnson:

dzenanz (approved):

PR #898, opened 2022-03-18
Remove deprecated configuration files for modern C++ and cmake

hjmjohnson:
STYLE: Remove deprecated compiler feature detection mechansims

Since cmake version 3.0 the use of The WriteCompilerDetectionHeader
module is not recommended.

Run "cmake --help-policy CMP0120" for policy details.

Prevent configuration time warning messages when using
cmake versions greater than 3.20

Preserve deprecation specialization for c++11 in vcl_compiler.h


hjmjohnson:

vcl/vcl_compiler_detection.h
```diff
@@ -1,314 +1,22 @@
...
-# define VXL_COMPILER_IS_Comeau 1
+Run "cmake --help-policy CMP0120" for policy details.
 
-#elif defined(__INTEL_COMPILER) || defined(__ICC)
-# undef VXL_COMPILER_IS_Intel
-# define VXL_COMPILER_IS_Intel 1
+Prevent configuration time warning messages when using
+cmake versions greater than 3.20
+*/
+#warning "The vcl_compiler_detection.h header file is no longer used.\n"   \
```
dzenanz, line 15: This is likely to produce a compile error on MSVC. See https://stackoverflow.com/questions/29694305/preprocessor-warning-equivalent-in-visual-c.

dzenanz (approved):

vcl/vcl_compiler_detection.h
hjmjohnson: @dzenanz   I will put conditional code for MSVC around this before I merge.  Thank you.

PR #902, opened 2022-08-19
STYLE: Remove unused cmake config variables

hjmjohnson:
Several variables were never possible to have set with VXL.  The values
were never set in any VXL codebase.

There are no find_package(Python* ) or other cmake "set" methods
for the values being propogated int the config files.

Remove the empty config settings.


hjmjohnson:
@dzenanz  This will clean up ITK a little bit.

dzenanz (approved):

PR #904, opened 2022-09-26
Trifocal upgrade

mundyj:
## PR Checklist

 :no_entry_sign: --> Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
- 
 :no_entry_sign: --> Makes changes to the contributed directory API DOES NOT require semantic versioning increase
-  Adds Documentation.


mundyj:
Shouldn't break any interfaces just an additional set function - none removed.  Updated the version number anyway

hjmjohnson (approved):

PR #909, opened 2022-11-14
ENH: Updates to lbfgsb.c to support better diagnostics

hjmjohnson:
The failure mode of lbfgsb can be more specific than was previously allowed.


hjmjohnson:
@dzenanz   This is the vxl change that I missed that will make ITK provide better diagnostics.

dzenanz (approved):

PR #910, opened 2022-11-22
Large disparity

mundyj:
## PR Checklist

:no_entry_sign: --> Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase
 :no_entry_sign: --> Makes design changes to existing vxl/core\* API that requires semantic versioning increase

 :no_entry_sign: --> Adds tests and baseline comparison (quantitative).
:no_entry_sign: --> Adds Documentation.

### DIFF
--- a/contrib/brl/bseg/bsgm/bsgm_error_checking.hxx
+++ b/contrib/brl/bseg/bsgm/bsgm_error_checking.hxx
@@ -399,7 +399,7 @@ void bsgm_compute_invalid_map(
     img_stop_x = target_window.max_x();
     img_stop_y = target_window.max_y();
   }
-
+  
   // the x coordinate of the left-most possible reference pixel mapped to from the target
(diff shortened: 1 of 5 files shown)


mundyj:
bsgm was not searching over a large enough disparity range and required per-pixel initial guess as to the minimum start to the search. The existing capability for dense min-disparity is now working successfully. Also a function was added that finds the min and max disparity limits for a specified min and max height. The limits are found using the target and reference cameras

drewgilliam (changes requested):
Many debug statements that save to hardcoded windows paths are still present, please remove.

hjmjohnson:
@mundyj   Please squash the commits to that bugs are not added and removed in a single patch.

drewgilliam (approved):
@hjmjohnson - @mundyj has squashed, should be good

PR #916, opened 2023-01-26
COMP: Do not contaminate CMAKE_REQUIRED_FLAGS

hjmjohnson:
VXL contaminates the CMAKE_REQUIRED_FLAGS that may be used by other packages.  This is only relavant for compilations on i386 machines. As stated in the gcc docs the SSE instruction set is enabled by default when using an x86-64 compiler.

Long-standing problem identified in https://github.com/InsightSoftwareConsortium/ITK/pull/3852


hjmjohnson:
@dzenanz  Upstream fix for https://github.com/InsightSoftwareConsortium/ITK/pull/3852

dzenanz (approved):

thewtex (approved):

PR #926, opened 2023-08-30
PERF: Speed up `vnl_sparse_matrix::operator()`

N-Dekker:
Made access to the back of a row _O(1)_ (constant-time), rather than _O(n)_. Simplified the loop to find column `c` (in case it is not in the back).

Adding 1000+ columns to the end appears to become more than 5 times as fast. Retrieving existing values from any column appears to become slightly faster than before. Observed using Visual Studio 2022, Release.

### DIFF
--- a/core/vnl/vnl_sparse_matrix.hxx
+++ b/core/vnl/vnl_sparse_matrix.hxx
@@ -413,16 +413,28 @@ T& vnl_sparse_matrix<T>::operator()(unsigned int r, unsigned int c)
 {
   assert((r < rows()) && (c < columns()));
   row& rw = elements[r];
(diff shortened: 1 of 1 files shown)


decrispell (approved):

N-Dekker:
@decrispell Thanks for your approval, Daniel! Note that similar code changes could be applied to three other `vnl_sparse_matrix` member functions as well: the `const` overload of `operator()`, the `get`, and the `put` member function. For `put`, I would also expect quite some performance improvement, when it can do `push_back` without having to iterate over all row elements. 

Anyway, that might be something to consider for a follow-up. 😃

PR #942, opened 2024-02-28
STYLE: `vnl_matrix_fixed::is_equal` assume equal number of rows, columns

N-Dekker:
The check that the corresponding number of rows and columns are equal appears redundant, for `vnl_matrix_fixed`.

### DIFF
--- a/core/vnl/vnl_matrix_fixed.hxx
+++ b/core/vnl/vnl_matrix_fixed.hxx
@@ -699,9 +699,6 @@ bool vnl_matrix_fixed<T,nrows,ncols>
   if (this == &rhs)                                      // same object => equal.
     return true;
 
-  if (this->rows() != rhs.rows() || this->cols() != rhs.cols())
-    return false;                                        // different sizes => not equal.
-
   for (unsigned int i = 0; i < nrows; ++i)
     for (unsigned int j = 0; j < ncols; ++j)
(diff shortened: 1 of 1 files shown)


core/vnl/vnl_matrix_fixed.hxx
```diff
@@ -699,9 +699,6 @@ bool vnl_matrix_fixed<T,nrows,ncols>
   if (this == &rhs)                                      // same object => equal.
     return true;
 
-  if (this->rows() != rhs.rows() || this->cols() != rhs.cols())
```
dzenanz, line 702: Maybe replace it by an assert, instead of removing entirely?

N-Dekker: Thanks @dzenanz, I thought so too, originally. But it just does not seem to have any added value. Both arguments are of the same type, `vnl_matrix_fixed<T,nrows,ncols>`. So _by definition_ they always have the same dimensions.

Of course you could add a `static_assert(rows() == nrows && cols() == ncols)`, but then you could do so in each and every member function. `vnl_matrix_fixed::rows()` and `vnl_matrix_fixed::cols()` are both constexpr, they will always just return the same value, for a specific `vnl_matrix_fixed<T,nrows,ncols>` instance.

PR #943, opened 2024-02-28
BUG: Fix `is_equal` support when vnl_vector or matrix contains NaN

N-Dekker:
It appears that `is_equal` would sometimes return `true` when a vnl_vector or
matrix would contain the floating point value NaN. Fixed by adjusting the
condition for doing a `return false` out of the `for` loop, in `is_equal`.

Added unit tests that compare a vnl_vector or matrix that contains NaN and one
that just contains the value zero.

So with this commit, `{ NaN, ..., NaN } is_equal { 0, ..., 0 }` will return _false_, as it should.

### DIFF
--- a/core/vnl/tests/test_matrix_fixed.cxx
+++ b/core/vnl/tests/test_matrix_fixed.cxx
@@ -407,6 +407,12 @@ test_float()
   TEST("m.update([4],1,1)",
        ((m3 = 4), (m.update(m3, 1, 1)), (m(0, 0) == 0 && m(0, 1) == -2 && m(1, 0) == 2 && m(1, 1) == 4)),
        true);
+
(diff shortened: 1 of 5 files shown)


N-Dekker:
Test failure at https://app.circleci.com/pipelines/github/vxl/vxl/134/workflows/538e0771-6b30-471d-8d70-986d07ec2c76/jobs/133:

```
99% tests passed, 1 tests failed out of 435

Total Test time (real) =  15.00 sec

The following tests FAILED:
        143 - vnl_algo_test_symmetric_eigensystem (Failed)
```

Details:
```
Fixed Time: 71365168   netlib time: 28670065
 Test 009: Specialised version is faster                         --> **FAILED**

```

Sorry, I don't really understand how that test failure relates to my pull request 🤷

decrispell (approved):

PR #954, opened 2024-07-05
COMP: Replacement of sprinf to snprintf (trivial cases)

andrei-sandor:
Hi, I will create some PRs to fix clang warnings for the vxl repo.

Here, it is a replacement of sprintf to snprintf for easy cases where an array with a specific size is involved (sizeof of that array).


seanm:
@hjmjohnson care to review?

core/vil1/vil1_stream_url.cxx
```diff
@@ -226,10 +226,10 @@ vil1_stream_url::vil1_stream_url(char const * url)
   char buffer[4096];
 
   // send HTTP 1.1 request.
-  std::sprintf(buffer, "GET /%s / HTTP/1.1\n", path.c_str());
+  std::snprintf(buffer, sizeof(buffer), "GET /%s / HTTP/1.1\n", path.c_str());
   if (!auth.empty())
-    std::sprintf(buffer + std::strlen(buffer), "Authorization:  Basic %s\n", encode_base64(auth).c_str());
-    //    std::sprintf(buffer+std::strlen(buffer), "Authorization:  user  testuser:testuser\n");
+    std::snprintf(buffer + std::strlen(buffer), sizeof(buffer), "Authorization:  Basic %s\n", encode_base64(auth).c_str());
+    //    std::snprintf(buffer+std::strlen(buffer), sizeof(buffer) "Authorization:  user  testuser:testuser\n");
```
seanm, line 232: @andrei-sandor this isn't quite right... notice the 2nd call does not start at the address of `buffer` but `buffer` plus an offset, therefore the size remaining is less than originally; less by the amount already written.

I think you want:

`std::snprintf(buffer + std::strlen(buffer), sizeof(buffer) - std::strlen(buffer), ...`

Maybe make sure I'm not off-by-1 there by sketching out an example on paper...

core/vil1/vil1_stream_url.cxx
```diff
@@ -226,10 +226,10 @@ vil1_stream_url::vil1_stream_url(char const * url)
   char buffer[4096];
 
   // send HTTP 1.1 request.
-  std::sprintf(buffer, "GET /%s / HTTP/1.1\n", path.c_str());
+  std::snprintf(buffer, sizeof(buffer), "GET /%s / HTTP/1.1\n", path.c_str());
   if (!auth.empty())
-    std::sprintf(buffer + std::strlen(buffer), "Authorization:  Basic %s\n", encode_base64(auth).c_str());
-    //    std::sprintf(buffer+std::strlen(buffer), "Authorization:  user  testuser:testuser\n");
+    std::snprintf(buffer + std::strlen(buffer), sizeof(buffer) - std::strlen(buffer) - 1 , "Authorization:  Basic %s\n", encode_base64(auth).c_str());
+    //    std::snprintf(buffer+std::strlen(buffer), sizeof(buffer) "Authorization:  user  testuser:testuser\n");
```
seanm, line 232: Looking at this again, it's not clear 4096 is even big enough, so then you'd have to check the return value of snprintf, and then it gets all ugly.

I'd remove this from this 'trivial' PR and rewrite this bit to use `std::string`...

seanm (approved):

seanm:
@decrispell care to review?

PR #956, opened 2024-07-05
 COMP: Replace sprintf to snprintf

andrei-sandor:
Replacement of sprintf to snprintf.

Here only one fix is proposed since it involves doing more involved changes.

Here the snprintf requires the size of buf. Unfortunately, buf, without it’s length, is passed to the function.

We can also see that each function in the file has the same name and a argument list very similar for each of them.

I propose to create at the top a function that has the same name as the others, but with the length of buf added and snprintf used. Then, use that function for the other functions accordingly.

Can I proceed like this for the rest of the functions in the file? Also, should we add a deprecated comment?

### DIFF
--- a/core/vnl/vnl_matlab_print_scalar.cxx
+++ b/core/vnl/vnl_matlab_print_scalar.cxx
@@ -4,22 +4,39 @@
 #include <cstdlib>
 #include <cstring>
 #include <complex>
+
 #include "vnl_matlab_print_scalar.h"
+#include "vcl_compiler.h"
+
 
 void
-vnl_matlab_print_scalar(int v, char * buf, vnl_matlab_print_format)
+vnl_matlab_print_scalar(int v, char * buf, size_t buf_len, vnl_matlab_print_format)
 {
-  std::sprintf(buf, "%4d ", v);
+  std::snprintf(buf, buf_len, "%4d ", v);
 }
 
 void
-vnl_matlab_print_scalar(unsigned v, char * buf, vnl_matlab_print_format)
+vnl_matlab_print_scalar(int v, char * buf, vnl_matlab_print_format) 
 {
-  std::sprintf(buf, "%4u ", v);
+ vnl_matlab_print_scalar(v, buf, SIZE_MAX);
+}
+
+
+void
+vnl_matlab_print_scalar(unsigned v, char * buf, size_t buf_len, vnl_matlab_print_format)
+{
+  std::snprintf(buf, buf_len, "%4u ", v);
 }
 
 void
-vnl_matlab_print_scalar(float v, char * buf, vnl_matlab_print_format format)
+vnl_matlab_print_scalar(unsigned v, char * buf, vnl_matlab_print_format) 
+{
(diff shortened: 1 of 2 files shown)


seanm:
@decrispell @hjmjohnson So is it ok for us to proceed in this fashion, and change function signatures where necessary to bring along the buffer length?

core/vnl/vnl_matlab_print_scalar.cxx
```diff
@@ -4,88 +4,131 @@
...
 #include <cstring>
 #include <complex>
+
 #include "vnl_matlab_print_scalar.h"
+#include "vcl_compiler.h"
+
 
 void
-vnl_matlab_print_scalar(int v, char * buf, vnl_matlab_print_format)
+vnl_matlab_print_scalar(int v, char * buf, size_t buf_len, vnl_matlab_print_format format)
```
seanm, line 13: You adding the parameter name `format` here might introduce a compiler warning about "format is unused". A common way to suppress such warnings is to remove the parameter name, which is probably why it was omitted. Not sure why you added it, but probably should remove it.  Not just on this line, but elsewhere below too.

core/vnl/vnl_matlab_print_scalar.cxx
```diff
@@ -130,38 +173,41 @@ vnl_matlab_print_scalar(std::complex<double> v, char * buf, vnl_matlab_print_for
...
   if (r == 0)
   {
-    std::sprintf(fmt,
-                 "%%"
-                 "%d"
-                 "d ",
-                 width);
-    std::sprintf(buf, fmt, 0);
+    std::snprintf(fmt,
+                  buf_len,
```
seanm, line 177: Here the 2nd parameter is wrong. Not matching the right buffer. Elsewhere too.

core/vnl/vnl_matlab_print_scalar.h
```diff
@@ -32,6 +33,34 @@ vnl_matlab_print_scalar_declare(std::complex<float>);
...
 vnl_matlab_print_scalar_declare(std::complex<long double>);
 
+
+
+
+
+
+//: print real or complex scalar into character buffer.
+#define vnl_matlab_print_scalar_declare_old(T) \
+VNL_EXPORT void vnl_matlab_print_scalar(T v, \
```
seanm, line 43: I think this could have `VXL_DEPRECATED` too, ie:

`VNL_EXPORT VXL_DEPRECATED void vnl_matlab_print_scalar(T v, \`

Worth a try anyway.

seanm (changes requested):

seanm (approved):

seanm:
@hjmjohnson this lgtm now, care to review? @andrei-sandor ends his internship with us next week, so ideally we could wrap this up by then.

seanm:
@hjmjohnson friendly ping...

seanm:
@hjmjohnson friendly ping...

hjmjohnson (approved):

PR #958, opened 2024-07-25
 COMP: Replacement of sprinf by using std::string

andrei-sandor:
Replaced the sprintf with std::string. We didn't use snprint since it's not clear 4096 is even big enough to perform the operations. We have to check the return values of snprint which becomes more complex.

At the read from socket into memory, we decided to use the char buffer[4096] lower than where it was initially since std::string where we look at its size won't perform the same result as if char buffer[4096] is used.

### DIFF
--- a/core/vil1/vil1_stream_url.cxx
+++ b/core/vil1/vil1_stream_url.cxx
@@ -222,24 +222,20 @@ vil1_stream_url::vil1_stream_url(char const * url)
     return;
   }
 
-  // buffer for data transfers over socket.
-  char buffer[4096];
-
   // send HTTP 1.1 request.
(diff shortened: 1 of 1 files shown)


seanm:
This looks correct to me.

@hjmjohnson care to review?

decrispell:
Thanks - Out of curiosity, is this a problem that you encountered in use, or based on something uncovered by static code analysis / compiler / etc.?

decrispell:
> Thanks - Out of curiosity, is this a problem that you encountered in use, or based on something uncovered by static code analysis / compiler / etc.?

Nmd, I think my question is answered in https://github.com/vxl/vxl/pull/954

seanm:
Yes, "just" compiler warnings. OpenBSD and macOS unconditionally warn about using sprintf due to risk of buffer overflow.

PR #963, opened 2024-10-10
Pyramid image byte changes

eben0012:
ENH Update image pyramid to work with color images, byte and short images

@decrispell 

## PR Checklist
- [X] Makes breaking changes to the vxl/core/\* API that requires semantic versioning increase

### DIFF
--- a/core/vil/vil_pyramid_image_view.hxx
+++ b/core/vil/vil_pyramid_image_view.hxx
@@ -31,8 +31,8 @@ vil_pyramid_image_view<T>::vil_pyramid_image_view(vil_image_view_base_sptr image
     return;
 
   // it works with grey scale images
-  if (img->nplanes() != 1)
-    return;
+  //if (img->nplanes() != 1)
+   // return;
 
   images_.resize(levels);
   scales_.resize(levels);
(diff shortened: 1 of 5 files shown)


core/vil/vil_pyramid_image_view.hxx
```diff
@@ -30,9 +30,8 @@ vil_pyramid_image_view<T>::vil_pyramid_image_view(vil_image_view_base_sptr image
   if (!img)
     return;
 
-  // it works with grey scale images
-  if (img->nplanes() != 1)
-    return;
+  //if (img->nplanes() != 1)
+   // return;
```
decrispell, line 34: Remove unnecessary code unless there is some specific reason it's still useful as a comment

core/vpgl/vpgl_utm.h
```diff
@@ -56,7 +56,7 @@ class vpgl_utm
                double& x_out, double& y_out) const;
 
  private:
-   double a_{6378137}, b_{6356752.3142};
+   double a_{6378137}, b_{6356752.31424518};
```
decrispell, line 59: Did you mean to include this in this PR?  (Also, seems odd to increase precision of `b_` but not `a_`)

eben0012: My understanding is that we had some discrepancies between our calculations and other's and this precision increase was the fix to correct that. It could be put into a separate PR if necessary

decrispell: Ideally unrelated changes should go in separate PRs but this seems innocuous enough..

decrispell (approved):

PR #966, opened 2024-11-26
Use cmake_minimum_required to set policy.

blowekamp:
Add "...policy_max", sets the maximum policy, and avoid CMake deprecation warnings when the min version is too old.

## PR Checklist

-  Adds Documentation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,11 +13,9 @@
 #
 #   vxl-maintainers@lists.sf.net
 
-cmake_minimum_required(VERSION 3.10.2)
+cmake_minimum_required(VERSION 3.10.2...3.20.3)
 if(CMAKE_VERSION VERSION_LESS 3.12.0)
   cmake_policy(VERSION ${CMAKE_VERSION})
-else()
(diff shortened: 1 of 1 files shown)


blowekamp:
@hjmjohnson Can you look at this?

blowekamp:
@dzenanz Can you please review?

dzenanz (changes requested):
What happens with CMake versions between 3.10.2 and 3.12?

The existing logic seems to better cover these possibilities.

blowekamp:
It this same. From the documentation:

https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html
```
If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.
```

dzenanz (approved):

blowekamp:
Can this PR be merged?

decrispell:
> Can this PR be merged?

I was waiting on @hjmjohnson since you requested his review as well, but I'll go ahead and merge now.

PR #970, opened 2024-12-20
Clang tidy modernizations

hjmjohnson:
Implements detection of local variables which could be declared as const but
    are not. Declaring variables as const is recommended by many coding
    guidelines, such as: ES.25 from the C++ Core Guidelines.
    
    cd ${BLDDIR}
    run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,misc-const-correctness -header-filter=.* -fix
    
    The emptiness of a container should be checked using the empty() method
    instead of the size() method. It is not guaranteed that size() is a
    constant-time function, and it is generally more efficient and also
    shows clearer intent to use empty(). Furthermore some containers may
    implement the empty() method but not implement the size() method. Using
    empty() whenever possible makes it easier to switch to another container
    in the future.
    
    cd ${BLDDIR}
    run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,readability-container-size-empty  -header-filter=.* -fix
    
    Replaces explicit calls to the constructor in a return with a braced
    initializer list. This way the return type is not needlessly duplicated in the
    function definition and the return statement.
    
    The check flags insertions to an STL-style container done by calling the
    push_back method with an explicitly-constructed temporary of the container
    element type. In this case, the corresponding emplace_back method results in
    less verbose and potentially more efficient code.
    
    This check replaces default bodies of special member functions with
    = default;. The explicitly defaulted function declarations enable more
    opportunities in optimization, because the compiler might treat
    explicitly defaulted functions as trivial.
    
    Additionally, the C++11 use of = default more clearly expreses the
    intent for the special member functions.
    
    This check replaces undefined special member functions with
    = delete;. The explicitly deleted function declarations enable more
    opportunities in optimization, because the compiler might treat
    explicitly deleted functions as noops.
    
    Additionally, the C++11 use of = delete more clearly expresses the
    intent for the special member functions.
    
    Converts a default constructor’s member initializers into the new
    default member initializers in C++11. Other member initializers that match the
    default member initializer are removed. This can reduce repeated code or allow
    use of ‘= default’.


hjmjohnson:
@dzenanz This implements some of the same modernization refactorings that have been successful in ITK to the VNL core.

core/vnl/algo/tests/test_symmetric_eigensystem.cxx
```diff
@@ -92,63 +92,63 @@ test_symmetric_eigensystem()
   { // compare speed and values of specialised 3x3 version with nxn version
     constexpr unsigned n = 20000;
 
+    // Generate a random system with seeded value to known value
```
dzenanz, line 95: I did not actually review this refactoring commit. Possibly remove `# Please enter the commit message ...` from commit message.

vcl/internal/vcl_atomic_count_sync.h
```diff
@@ -44,9 +44,9 @@ class vcl_atomic_count
   operator long() const { return __sync_fetch_and_add(&value_, 0); }
 
 private:
-  vcl_atomic_count(const vcl_atomic_count &);
+  vcl_atomic_count(const vcl_atomic_count &) = delete;
```
dzenanz, line 47: Removal of corresponding implementations should be part of this commit.

dzenanz (approved):
Mostly looks good.

vcl/internal/vcl_atomic_count_sync.h
hjmjohnson: There is no implementation for these private copy constructors since they were never allowed to be used.

core/vnl/algo/tests/test_symmetric_eigensystem.cxx
hjmjohnson: Done.

dzenanz (approved):

PR #981, opened 2025-05-21
Extend rsm metdata

mundyj:
## PR Checklist

 :no_entry_sign: --> Makes design changes to existing vxl/core\* API that requires semantic versioning increase

 :no_entry_sign: --> Makes changes to the contributed directory API DOES NOT require semantic versioning increase
 :no_entry_sign: --> Adds tests and baseline comparison (quantitative).
:no_entry_sign: --> Adds Documentation.

### DIFF
--- a/core/vpgl/file_formats/vpgl_nitf_RSM_camera_extractor.cxx
+++ b/core/vpgl/file_formats/vpgl_nitf_RSM_camera_extractor.cxx
@@ -295,7 +295,9 @@ bool vpgl_nitf_RSM_camera_extractor::process_polytope(size_t image_subheader_ind
   meta.bounding_box_valid = true;
 
   // polytope vertex indices at image corners
-  unsigned UL = 3, UR = 4, LL = 1, LR = 2;
+  // doesn't seem to have a fixed relation to
+  // polytope vertices in the test examples.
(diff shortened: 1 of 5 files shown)


core/vpgl/file_formats/vpgl_nitf_RSM_camera_extractor.cxx
```diff
@@ -614,8 +680,10 @@ vpgl_nitf_RSM_camera_extractor::scan_for_RSM_data(bool verbose)
 
         nitf_tre<std::string> nt3("SID", false, true);
         nt3.get_append(tres_itr, tre_str, v);
+        if(nt3.get(tres_itr, mdata.sid_))
+           mdata.sid_valid = true;
 
-        nitf_tre<std::string> nt4("STID", false, true);
+           nitf_tre<std::string> nt4("STID", false, true);
```
decrispell, line 686: is this line intended to be part of the `if` block?  (Need braces if so, added intentation is misleading if not)

mundyj:
If the metadata mode is zero_to_360  code "H" we want to convert to +-180 thus the mismatch it is doing what was intended

The indent was a typo so the right thing was happening but I lined it up

decrispell (approved):

Issue #983, opened 2025-06-03
vxl/CMakeLists.txt has obsolete list of C / C++ standards, I propose just remove them


seanm:
vxl/CMakeLists.txt has logic like:

```
set(VALID_C_STANDARDS "90" "99" "11")
if(NOT CMAKE_C_STANDARD IN_LIST VALID_C_STANDARDS )
   MESSAGE(FATAL_ERROR "CMAKE_C_STANDARD:STRING=${CMAKE_C_STANDARD} not in know standards list\n ${VALID_C_STANDARDS}.")
endif()
```

Since C11 there has been C17 and C23.  I've upped all my builds to C17 and now get an error:

```
CMake Error at Modules/ThirdParty/VNL/src/vxl/CMakeLists.txt:46 (MESSAGE):
  CMAKE_C_STANDARD:STRING=17 not in know standards list

   90;99;11.
```

Similar for C++.

Seems to me all these checks should just be removed. There will always be a newer language dialect in the future that we can't list in the present, and trying to preflight them will just cause this same problem again in the future.

What's the point of these checks?

seanm:
@hjmjohnson @dzenanz maybe you guys have thoughts on this?  It's blocking my switch from C11 to C17 as it breaks the ITK build.

seanm:
>I believe the point of these checks is that the user does not accidently start building with a C++ version which is known not to work.

Personally, I'd not a fan of this whitelist approach, because we'll just have this problem again and again as more C and C++ revisions occur, which they will.

Deliberately breaking the build like this hurts bisecting and other things.

But I can live with it. :)

PR #984, opened 2025-06-04
ENH: Update valid C language levels

hjmjohnson:
Resolves: #983 to allow building with C17 or C23


seanm:
I'm just trying exactly this locally.... will get back to you...

seanm:
So it builds as C17, but not as C23. Working on a patch now...

PR #985, opened 2025-06-04
Fixed compilation errors with C23, and one with C++

seanm:


hjmjohnson:
@seanm Thanks for your comments. I updated to allow building with c23.

v3p/netlib/tests/tricall.c
```diff
@@ -144,7 +138,7 @@ int reportnorms;
 /*                                                                           */
 /*****************************************************************************/
 
-int main()
+int main(void)
```
dzenanz, line 141: Is this change needed? I thought that `main(void)` is the **old** style?

dzenanz (approved):
LGTM

PR #990, opened 2025-09-03
update libf2c

hjmjohnson:
ENH: Update libf2c to March 2024 revision

- **ENH: Added script to assist with updating libf2c**
- **ENH: Update libf2c to the March 2024 revision**

## PR Checklist

- :no_entry_sign:  Makes design changes to existing vxl/core\* API that requires semantic versioning increase
- [X] Makes changes to the contributed directory API DOES NOT require semantic versioning increase


hjmjohnson:
@dzenanz This is a low priority, but I would appreciate your comments.

dzenanz:
If this works, it is fine by me. Have you thought of automating these updates, similarly to ITK's UpdateFromUpstream infrastructure?

hjmjohnson:
Yes, but the updates only happen 1x every 15 years :)... so the ROI would take a millennium :).

PR #996, opened 2025-10-25
COMP: Remove rpoly-related code due to ACM copyright restrictions

hjmjohnson:
The vxl core and it's dependancies should not depend on
restrictive copyrights.


hjmjohnson:
> @hjmjohnson The changes look fine to me, but I think the VXL major version number (in the root CMakeLists.txt) should be increased because of the change to the core API (removal of vnl/algo/vnl_rpoly_roots)

Will do! Thanks!

PR #1037, opened 2026-05-04
Raise C++ baseline to 17 and demonstrate vnl maintenance cleanups (stacked on #1036)

hjmjohnson:
I would like community feedback regarding updating VXL to CXX_STANDARD=17 language conformance.  ITK, Slicer, and many other (medical imaging downstream) packages already require c++17.  Please respond to comments on this PR regarding support/concerns for moving to a c++17 baseline language requirement.

**WIP / DRAFT — stacked on top of #1036.** Do not merge before #1036.

Raises the vxl C++ baseline to C++17 (was C++11) and demonstrates a focused set of maintenance/style cleanups that the new floor unlocks. Each commit builds + passes the full ctest suite locally (692/692) with `VXL_BUILD_CONTRIB=ON`. No `contrib/` files are touched.

<details>
<summary>Commits</summary>

1. `BUILD: Raise minimum C++ standard to C++17 and bump VXL to 7.0.0.0` — flips the `CMAKE_CXX_STANDARD` default from `11` to `17`. Major version bump (6.3.0.0 → 7.0.0.0) signals the toolchain requirement to consumers.
2. `STYLE: Use 'inline constexpr' for vnl_math header constants` — converts the 24 namespace-scope constants in `vnl_math.h` from `static constexpr` (per-TU internal-linkage copies) to `inline constexpr` (single program-wide object). Removes the obsolete pre-C++17 rationale comment.
3. `STYLE: Use C++17 _v / _t aliases and 'if constexpr' in vnl_error.h` — collapses `typename std::enable_if<...>::type = 0` to `std::enable_if_t<..., int> = 0`, factors the four range-check assertion bodies into a single `vnl_error_assert_int_range` helper, and uses `if constexpr` to elide the assertions at compile time when the source type is provably representable in `int`.
4. `STYLE: Collapse vnl_complex_traits specializations with SFINAE` — the `VCL_DEFINE_SPECIALIZATION_MACRO` X-macro and 14 explicit specializations are replaced by 3 partial specializations keyed on `std::is_integral_v` / `std::is_floating_point_v`. Net: −119/+23 lines.
5. `STYLE: Add [[nodiscard]] to vnl predicate / size accessors` — 24 accessor declarations on `vnl_matrix`, `vnl_vector`, `vnl_matrix_fixed`, `vnl_vector_fixed`: `is_zero`, `is_finite`, `empty`, `size`, `rows`, `cols`, `columns`. Compiler now warns when the result is discarded.
6. `STYLE: Replace 'typedef' with 'using' across core/vnl` — mechanical conversion of 224 `typedef` declarations in 71 files to the C++11+ `using Name = Type;` form. Excludes `typedef enum` / `typedef struct` / `typedef union` (different syntax, two cases).
7. `STYLE: Use structured bindings for sparse_vector pair iteration` — five `for (auto & itr : sparse_vector) { const int j = itr.second; const int k = itr.first; ... }` loops in `vnl_sparse_lm.cxx` collapse to `for (auto & [k, j] : sparse_vector) { ... }`.

</details>

<details>
<summary>Tier-3 items investigated but not applied</summary>

- **Fold expressions.** Variadic templates aren't a vnl idiom; no opportunities found in `core/vnl`.
- **Structured bindings beyond `vnl_sparse_lm.cxx`.** vnl public APIs return single objects; only `std::pair`-iteration pattern in the sparse Levenberg-Marquardt code benefits.

</details>

<details>
<summary>What this PR is NOT</summary>

- Not a performance change. C++17 buys readability and ODR cleanup, not runtime speed; vnl's hot paths are BLAS / netlib / inner kernel loops where the language version doesn't matter.
- Not a contrib touch — explicitly out of scope.
- Not exhaustive. Additional C++17 wins for follow-up PRs:
  - `std::optional` returns from `vnl_matrix_inverse` / `vnl_qr::solve` / `vnl_svd::solve` (API-breaking — needs v8.0).
  - `std::string_view` in `vnl_error` / `vnl_matlab_print_scalar` signatures.
  - Collapse `vnl_numeric_traits` specializations the same way commit 4 did for `vnl_complex_traits`.
  - More `constexpr` on `vnl_vector_fixed` / `vnl_matrix_fixed` constructors and `vnl_math` arithmetic helpers.

`std::numbers::pi` is **C++20**, not C++17 — that delegation has to wait for a future floor bump.

</details>

<details>
<summary>Local validation</summary>

```
cmake --build build         # full tree, contrib on
ctest --test-dir build -j8  # 692/692 pass, every commit
pre-commit run              # clean for touched files
```

Validated at each of the seven commits, not just the tip.

</details>


hjmjohnson:
A note on the maintenance / developer-burden case for the C++17 floor, and why **now** (as of 2026-05) is the right time. The bump primarily *removes* friction rather than adding capability — vxl's documented C++11 floor is already a paper floor (CI builds at C++17 on Ubuntu/macOS/Windows; the prior CircleCI-only C++11 leg is fixed in commit 1 of this PR). What changes is what contributors no longer have to think about.

<details>
<summary>Why now is the right time</summary>

Several conditions have converged that make 2026 the lowest-cost moment to flip the floor; waiting incurs cost without buying anything.

1. **CI already validates C++17.** The GitHub Actions matrix (Ubuntu-24.04 + macOS-14 + Windows-2022, both `itk-core` and `full-modules` legs) builds the entire tree — including contrib under `VXL_BUILD_NONDEPRECATED_ONLY=ON` — at `CMAKE_CXX_STANDARD=17`. 
2. **Compiler floor for C++17 is now firmly behind us.** GCC 7 (May 2017), Clang 5 (Sept 2017), VS 2017 15.7 (May 2018), Apple Clang in Xcode 9.3 (March 2018). All four are 7+ years old. GCC 6 EOL'd in 2019; VS 2015 mainstream support ended 2020. Distros that vxl actively targets (Ubuntu LTS, RHEL, Debian stable, Homebrew) all ship compilers far past C++17-complete.
3. **Many current compilers default to C++17 and require opt-*out* to use older versions.** GCC 11+ defaults to `-std=gnu++17`; Clang 16+ defaults to C++17; current Apple Clang (Xcode ≥ 14) defaults to C++17. A user invoking `gcc` / `clang` / `clang++` with no `-std=` flag is *already* compiling vxl under C++17 today — vxl's `CMAKE_CXX_STANDARD=11` only takes effect because CMake explicitly emits `-std=c++11`. In other words: when downstream consumers stop passing `-DCMAKE_CXX_STANDARD=…` (which they often do when integrating via `find_package(VXL)`), they get C++17 *by default* from the toolchain. The documented C++11 floor doesn't reflect what the build is actually doing in practice. (MSVC remains the lone holdout — defaults to C++14 unless `/std:c++17` is set — but every CI leg already passes `/std:c++17` or its CMake equivalent.)
4. **A principal downstream consumer (ITK ≥ 5.4) already requires C++17.** Anyone consuming vxl through ITK is paying the C++17 cost regardless. Bumping vxl's own floor doesn't increase their burden — it ends the awkward situation where vxl claims to support older toolchains than its largest consumer does.
5. **Modernization momentum is high.** 80 STYLE/COMP/ENH commits in the last 12 months on master (recent: vnl_random C-array→`std::array`, `lrand32` deprecation, `std::copy_n` / `std::fill_n` / `std::accumulate` substitutions, `std::numeric_limits` displacing `INT_MAX`/`INT_MIN`). Each of those PRs carried an implicit "does this work at C++11?" judgement call. Doing the floor bump now amortizes that question across the next year of cleanup PRs, instead of paying it 80 more times.
6. **Skipping to C++20 is not yet ready.** Full C++20 (concepts, coroutines, `<ranges>`, modules) requires GCC 11 / Clang 13 / MSVC 19.29 — first viable in 2022. As of 2026, those compilers are universal on dev workstations but not yet on every distro that vxl is packaged for. C++17 is the current sweet spot: meaningful feature wins (`if constexpr`, `_v`/`_t`, `inline constexpr`, `std::optional`, `std::string_view`, `[[nodiscard]]`) without excluding any actively-supported distro toolchain.
7. **Major version bump is already on the table.** vxl 6.3.0.0 → 7.0.0.0 in commit 1 is the standard signaling mechanism for "downstream packagers, please re-pin." Doing this bump, this version increment, and the visible cleanups together is one coordination event for the ecosystem rather than two.
8. **Latent C++17-incompatible code costs more the longer it sits.** `contrib/brl/bbas/imesh/imesh_operations.cxx`'s 2× `std::auto_ptr` and `contrib/mul/vil3d/tests/`'s `std::bind1st`/`bind2nd` are *removed* (not deprecated) in C++17. Any compiler in strict-C++17 mode rejects them. The longer this is the documented-floor's responsibility to support, the more contortions accumulate. Bumping the floor surfaces these cleanly as one-shot follow-up PRs.

Cost of waiting another release cycle: zero new capability gained, ~12 more "does this work at C++11?" judgement calls on follow-up PRs, continued documentation/reality drift, continued risk of a compiler-strict-mode bug report against `auto_ptr`/`bind*` that would have to be hot-patched in vxl 6.x.

</details>

<details>
<summary>Concrete maintenance wins (grounded in <code>core/</code> grep counts)</summary>

| Today's friction | C++17 relief |
|---|---|
| 23 `std::enable_if<…>::type = 0` SFINAE sites + 28 `std::is_xxx<T>::value` traits in `core/` | `enable_if_t` / `_v` aliases — half the keystrokes, dramatically shorter template-error messages |
| 84 `static constexpr` namespace-scope constants → per-TU internal-linkage copies (ODR foot-gun for taking addresses) | `inline constexpr` — single program-wide object, no per-TU bloat. Demo: `vnl_math.h`'s 24 constants in commit 2. |
| 0 `[[nodiscard]]` annotations on predicate accessors (`is_finite`, `is_zero`, `empty`, `size`, …) — silent bugs when caller forgets the comparison | Compile-time warning. Demo: commit 5 added 24 sites and CI immediately surfaced 8 latent test-coverage gaps in `test_container_interface.cxx`. |
| 4-overload tag-dispatch for "is source representable in `int`?" (`vnl_error.h`) | `if constexpr` collapses to one helper; runtime check elided when statically provable. Demo: commit 3. |
| 14 hand-written `vnl_complex_traits` specializations via X-macro | 3 partial specializations keyed on `is_integral_v` / `is_floating_point_v`. Demo: commit 4 — −119/+23 lines, same behavior. |
| 829 `typedef` declarations in `core/` — irregular syntax for function pointers, ambiguous parsing rules | Mechanical `using Name = Type;` conversion. Demo: commit 6 does 224 in `core/vnl/`; 605 remaining are future opportunistic cleanup. |

</details>

<details>
<summary>What contributors stop having to track</summary>

- **Version-conditional code.** `core/` currently has zero `#if __cplusplus` guards; that's good, but every cleanup PR has to mentally check "would this compile at C++11?" against a floor no CI leg actually proves. After the bump, the answer is just "C++17 only" and the question goes away.
- **Macro-shimming for portable deprecation.** `vcl_compiler.h.in`'s `VXL_DEPRECATED_MSG` already prefers `[[deprecated(MSG)]]` when available — the bump removes the fallback branches and the `#if !defined(_MSC_VER)`-style guards around them.
- **Removed-in-C++17 stdlib.** The bump forces `contrib/brl/bbas/imesh/imesh_operations.cxx`'s 2× `std::auto_ptr` and `contrib/mul/vil3d/tests/`'s `std::bind1st`/`bind2nd` to be cleaned up — these are latent build failures that the `full-modules` CI leg may already be tripping over silently. Better to fix once, post-bump, than carry them indefinitely.
- **`throw()` exception specs in `core/vil/vil_math.h`.** 24 sites; deprecated in C++11/14 and removed in C++17 in favor of `noexcept`. Mechanical follow-up.

</details>

<details>
<summary>Why this lowers the bar for future modernization PRs</summary>

The 80 STYLE/COMP/ENH commits on master in the last 12 months show vxl is in active modernization (recent vnl_random C-array→`std::array`, `lrand32` deprecation, `std::copy_n` substitutions). Every one of those commits had to make a "does this work at C++11?" judgement call. After the floor bump, the next round (`std::optional` returns from `vnl_matrix_inverse` / `vnl_qr::solve` / `vnl_svd::solve`, `std::string_view` in `vnl_error` / `vnl_matlab_print_scalar`, more `if constexpr` in templated hot paths, `std::numeric_limits` continuing to displace `INT_MAX`/`INT_MIN` macros) all go through without that check.

Net: the bump itself is small; the *compounding* effect on every subsequent maintenance PR is the larger win.

</details>

decrispell (approved):

decrispell:
Raising baseline to C++17 sounds good to me.

Issue #23, opened 2015-12-01
upgrade netlib to use LAPACK instead of LINPACK


dstoup:
vnl_algo uses the netlib library to provide good numerical algorithms.
During VNL's design, it was thought that the much newer LAPACK was better
than the LINPACK library. However, it was faster to get something working
with LINPACK, since it had less dependencies than LAPACK. Now, LAPACK's
superiority is undisputed. LAPACK is faster, and more accurate. It is
designed for efficient parallel processing including on SMP machines
so it should be thread safe. We do not know if f2c'ed LAPACK is
automally thread safe, but it should be possible to hand edit it
to deal with f2c's deficiencies. LAPACK3E has been specifically updated
to avoid SAVE statements and so is definitely thread safe. It is written
in Fortran 90 whilst f2c can still only read Fortran77. There is a
developmental version of gcc supporting Fortran95, so maybe gcc users
could optionally use the native LAPACK library.

netlib.org has a library called CLAPACK where someone else has done the
f2c and tidyup already. I don't know if it is threadsafe. It might not
be too much work to strip out any remaining static values.

It should be possible to start using LAPACK incrementally. LAPACK
names do not appear to clash with LINPACK, and since they both need
BLAS, adding the LAPACK code to v3p/netlib would be sensible.
I suggest starting with SVD of double matrices. This will require
also converting some helper functions. After that it should be
relatively straight forward to start importing individual LAPACK
routines, and moving vnl_algo classes over one at a time.

Note that vnl does not use BLAS, but does its own vector and matrix
multiplications. The speed of the MKL (An Intel supplied/optimised copy
of BLAS) was compared with gcc's compilation of VNL, and gcc/VNL won.
gcc/VNL was marginally faster on the matrixvector operations, although
the difference was more marked on the vectorvector operations.
However, LAPACK and LINPACK both need a BLAS, and the MKL 
was definitely faster than the gcccompiled version of the default BLAS
currently in netlib. So it would be worth providing an option to use an
external BLAS for vnl_algo/netlib. It might also be worth testing the
matrixmatrix operations  where MKL's knowledge of cache behaviour
may help more than vnl's knowledge of the matrix data organisation.
See https://sourceforge.net/mailarchive/message.php?msg_id=3628964 for
more details.

Some compiler optimsers (gcc's notably) appear to break bits of netlib.
Different versions of gcc (or even the same version of gcc on different
computers) break different bits of netlib, so turning optimisation off
[...]

PR #1041, opened 2026-05-08
BUG: Fix singleton races and serialize iterative-solver f2c paths (resolves #22)

### LINKED ISSUE #22: Ensure that VXL APIs are reentrant.
VXL is not threadsafe, but should be. Some of this is due to the
limitations of f2c in v3p/netlib. It produces static variables, where
automatic ones would do. This may be fixed with the latest version of
f2c and or lapack3e.

In other places, such as the vil image loader, and the vsl binary
loader, there are singleton classes. This code is inherently
nonthread safe. Such code can be clearly documented to indicate how
to avoid thread problems, e.g. by initialising such loaders in a single
thread.

hjmjohnson:
Fixes the singleton races flagged in issue #22 (vil image-loader, vsl binary-loader) plus a sibling race in vsl_indent, and adds a band-aid mutex around ARPACK + L-BFGS-B until the proper netlib upgrade in #23 lands. New `core/vnl/algo/README_threading.md` documents which solver paths are now safe to call concurrently. 5 commits, 6 files, +318 / -59. Built and tested green on AppleClang 21 with `VXL_BUILD_CONTRIB=ON` (705/705 tests).

Resolves #22.

<details>
<summary>What was racing and how it's fixed</summary>

| Concern | Was | Now |
|---|---|---|
| `vil_file_format::add_file_format` mutates a shared `std::list` | unsynchronised | `std::mutex` guarding the mutator path |
| `vsl_binary_loader_base::register_this()` did `if (p == nullptr) p = new ...; p->push_back(...)` | TOCTOU on the null check + unsynchronised push_back | mutex bundled with the registry struct in one Meyers singleton (so destruction order is unified) |
| `vsl_indent` per-stream indent map | unsynchronised insertion of new ostream entries from the print path | mutex around find/insert; per-stream pointer access remains unsynchronised (single-threaded use of a per-thread vsl_b_*stream is the natural pattern) |
| `vnl_lbfgsb::minimize` calls `setulb_` (L-BFGS-B's f2c-translated SAVE'd-locals) | concurrent calls corrupt SAVE'd state | per-class `lbfgsb_call_mutex` held for the duration of one minimize() invocation |
| `vnl_sparse_symmetric_eigensystem::CalculateNPairs` calls `dnlaso_`/`dsaupd_`/`dseupd_` (ARPACK SAVE'd state) AND threads through a static `current_system` reverse-communication shim | concurrent calls corrupt both | per-class `sse_call_mutex` covering both concerns |

</details>

<details>
<summary>Important correction to the 2015 issue framing</summary>

The original issue text said *"VXL is not threadsafe... due to the limitations of f2c in v3p/netlib. It produces static variables..."*. Investigating the actual f2c output shows this is over-broad:

- The 165 `.c` files with file-scope `static` declarations in `v3p/netlib/` are **overwhelmingly read-only constants** that f2c emits for FORTRAN's pass-by-reference convention (`static integer c__1 = 1;` so a routine has an addressable `1` to pass). These never mutate and are naturally thread-safe.
- The actual mutable SAVE'd state lives in **iterative** solvers — ARPACK, L-BFGS-B — because their per-call solver history needs to persist across reverse-communication callback boundaries.
- **Dense LAPACK/BLAS** (`dgesvd_`, `dgemm_`, `dgetrf_`, …) is therefore reentrant in practice. `vnl_svd`, `vnl_qr`, `vnl_cholesky`, `vnl_lu_decomp`, `vnl_real_eigensystem`, `vnl_symmetric_eigensystem`, `vnl_levenberg_marquardt`, etc. can all be called concurrently on independent inputs without locking.

This is documented in `core/vnl/algo/README_threading.md` with a per-solver inventory.

The proper fix for the iterative-path band-aid is issue #23 (regenerate netlib via thread-safe LAPACK/ARPACK). Once that lands the band-aid mutex can be removed.

</details>

<details>
<summary>Commit ordering and detour</summary>

```
555ba83d36  DOC:   threading README
d2dc0845ea  BUG:   ARPACK + L-BFGS-B band-aid mutex
cca9981137  BUG:   vsl_indent shared-map race
888f404967  BUG:   vsl_binary_loader_base TOCTOU
3ac73cef2c  BUG:   vil_file_format add_file_format race
```

The vsl_binary_loader_base commit's first attempt put the mutex in a separate Meyers-singleton from the registry data. At program exit, the lazily-constructed mutex destructed before the registry, and the registry's destructor then locked dead memory → 38 test aborts with `mutex lock failed: Invalid argument`. Fixed by embedding the mutex as a struct member of the registry holder so they share lifetime. The detour and lesson are recorded in the commit message of `888f404967` so the next maintainer who tries to "simplify" by splitting them again won't repeat the bug.

</details>

<details>

Per-commit, on every commit in this PR:

```
cmake -G Ninja -S . -B build-contrib -DVXL_BUILD_CONTRIB=ON -DCMAKE_BUILD_TYPE=Release
ninja -C build-contrib -j10              # exit 0 every commit
ctest --test-dir build-contrib -j10      # 705/705 every commit (after rerunning the
                                         #   pre-existing clsfy_test_adaboost flake)
```

Tested on macOS 26.4 SDK with AppleClang 21.0.0.

</details>

### DIFF
--- a/core/vil/vil_file_format.cxx
+++ b/core/vil/vil_file_format.cxx
@@ -2,6 +2,7 @@
 //:
 // \file
 #include <cstdlib>
+#include <mutex>
 #include "vil_file_format.h"
 
 vil_file_format::~vil_file_format() = default;
@@ -159,10 +160,26 @@ struct vil_file_format_storage
   }
 };
 
+//: Mutex guarding mutations of the shared format list.
+//  The list itself is initialized once via the Meyers-singleton in
+//  all() (C++11 guarantees thread-safe initialization of function-
+//  local statics). This mutex serializes concurrent add_file_format
+//  calls from multiple threads. Iteration via all() after the
+//  registration phase is unsynchronized; callers that register
+//  formats from multiple threads must serialize their iteration
+//  externally.
+static std::mutex &
+vil_file_format_list_mutex()
+{
+  static std::mutex m;
+  return m;
+}
+
 //: The function will take ownership of ff;
 void
 vil_file_format::add_file_format(vil_file_format * ff)
 {
+  std::lock_guard<std::mutex> guard(vil_file_format_list_mutex());
   std::list<vil_file_format *> & l = all();
 
   // Always add runtime-loaded formats to the front to allow them to replace
--- a/core/vnl/algo/README_threading.md
+++ b/core/vnl/algo/README_threading.md
@@ -1,0 +1,138 @@
+# vnl_algo — threading and reentrancy guarantees
+
+This document inventories which `vnl_algo` solver paths are safe to
+call concurrently from multiple threads and which require external
+serialization. It accompanies the partial fixes for issue #22.
+
+## Summary
+
+- **Dense linear-algebra paths (`vnl_svd`, `vnl_qr`, `vnl_cholesky`,
+  `vnl_lu_decomp`, `vnl_matrix_inverse`, `vnl_determinant`, ...) are
+  effectively reentrant.** They call into f2c-translated LAPACK/BLAS
+  routines (`dgetrf_`, `dgemm_`, etc.) whose only file-scope `static`
+  storage is read-only constants emitted by f2c for FORTRAN's pass-by-
+  reference convention. Two threads can call these paths concurrently
+  on independent matrices without corruption.
+
+- **Iterative solver paths (`vnl_sparse_symmetric_eigensystem`,
+  `vnl_lbfgsb`) are internally serialized as of this version.** Each
+  carries an internal `std::mutex` that holds for the duration of one
+  `CalculateNPairs` / `minimize` invocation. Multi-threaded callers
+  observe correct results, but at the cost that two threads cannot
+  actually solve different systems in parallel — the second blocks
+  until the first finishes.
+
+- **Singletons in `vil_file_format` and `vsl_binary_loader_base` are
+  internally synchronized.** Concurrent registration of formats /
+  binary loaders from multiple threads is safe.
+
+- **Dense LAPACK error reporting (`xerbla_`)** prints to stderr from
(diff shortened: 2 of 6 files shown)


hjmjohnson:
Updated `735b804b75` (was `d2dc0845ea`) to move the `lbfgsb_call_mutex` `lock_guard` from inside the reverse-communication loop body to the top of `vnl_lbfgsb::minimize()`. The earlier per-`setulb_()` scope released the lock between iterations, letting a second `minimize()` call interleave `setulb_` invocations against the same f2c-translated SAVE'd state — the exact corruption window this band-aid is supposed to close, and the behavior `README_threading.md` already documents ("each `CalculateNPairs` / `minimize` invocation" is serialized).

Greptile review on ITK PR [InsightSoftwareConsortium/ITK#6251](https://github.com/InsightSoftwareConsortium/ITK/pull/6251#discussion_r3215452460) flagged the mismatch when the integration branch was synced into ITK. The new scope matches the `vnl_sparse_symmetric_eigensystem::CalculateNPairs` pattern (one `lock_guard` at function entry, held until return).

<details>
<summary>Diff</summary>

- `lock_guard` declared once at `minimize()` entry, before the driver setup
- Inner `{ lock_guard … }` block around the single `setulb_` call removed; `setulb_` indented out one level
- Anonymous-namespace comment for `lbfgsb_call_mutex()` rewritten to describe the full-loop scope (no longer claims the lock is "per `setulb_` call")

</details>

hjmjohnson:
Moving ITK efforts for VXL to https://github.com/InsightSoftwareConsortium/vxl/tree/for/itk-vxl-master.  This is primary so that the ITK development team can focus on the ITK-needed features of vxl and remove the rest.

Issue #816, opened 2020-12-04
Build errors dcistrma.h, dcfilefo.h


jpeter42:
I get the following Build errors:

In file included from .../vxl/core/vil/file_formats/vil_dicom_stream.cxx:5:

.../vxl/core/vil/file_formats/vil_dicom_stream.h:7:10: fatal error: dcistrma.h: No such file or directory

.../vxl/core/vil/file_formats/vil_dicom.cxx:30:12: fatal error: dcfilefo.h: No such file or directory

gl-yziquel:
I get that same issue on an Ubuntu 22. There however is a file `/usr/local/include/dcmtk/dcmdata/dcfilefo.h`that I installed by installing dcmtk from source. It just is not picked up properly by the `include <dcfilefo.h>`in `vxl/core/vil/file_formats/vil_dicom.cxx`.

Seems to me like a cmake setup issue.

Issue #769, opened 2020-03-26
error: allocating an object of abstract class type 'vil_dicom_stream_factory'


yurivict:
```
In file included from /wrkdirs/usr/ports/misc/vxl/work/vxl-2.0.2-682-g9b4d7496d4/core/vil/file_formats/vil_dicom_stream.cxx:5:
/wrkdirs/usr/ports/misc/vxl/work/vxl-2.0.2-682-g9b4d7496d4/core/vil/file_formats/vil_dicom_stream.h:44:16: error: allocating an object of abstract class type 'vil_dicom_stream_factory'
    return new vil_dicom_stream_factory(*this);
               ^
/usr/local/include/dcmtk/dcmdata/dcistrma.h:150:37: note: unimplemented pure virtual method 'ident' in 'vil_dicom_stream_factory'
  virtual DcmInputStreamFactoryType ident() const = 0;
                                    ^ 
/wrkdirs/usr/ports/misc/vxl/work/vxl-2.0.2-682-g9b4d7496d4/core/vil/file_formats/vil_dicom_stream.cxx:120:24: error: allocating an object of abstract class type 'vil_dicom_stream_producer'
  : DcmInputStream(new vil_dicom_stream_producer(vs))
                       ^
/usr/local/include/dcmtk/dcmdata/dcistrma.h:60:18: note: unimplemented pure virtual method 'eos' in 'vil_dicom_stream_producer'
  virtual OFBool eos() = 0;
                 ^
/usr/local/include/dcmtk/dcmdata/dcistrma.h:69:24: note: unimplemented pure virtual method 'avail' in 'vil_dicom_stream_producer' 
  virtual offile_off_t avail() = 0;  
                       ^
```

vxl-2.0.2-682-g9b4d7496d4
clang-8
FreeBSD 12

Issue #768, opened 2020-03-26
Fails to build when dcmtk-3.6.5 is present: fatal error: 'dcistrma.h' file not found


yurivict:
```
In file included from /wrkdirs/usr/ports/misc/vxl/work/vxl-2.0.2-682-g9b4d7496d4/core/vil/file_formats/vil_dicom_stream.cxx:5:
/wrkdirs/usr/ports/misc/vxl/work/vxl-2.0.2-682-g9b4d7496d4/core/vil/file_formats/vil_dicom_stream.h:7:10: fatal error: 'dcistrma.h' file not found
#include <dcistrma.h>
         ^~~~~~~~~~~~
1 error generated.
```

vxl-2.0.2-682
clang-8
FreeBSD 12

--------

Also, cmake complains when dcmtk isn't found: http://beefy10.nyi.freebsd.org/data/113i386-default/529158/logs/vxl-2.0.2.629.log

```
-- Please set DCMTK_DIR and re-run configure (missing: DCMTK_config_INCLUDE_DIR DCMTK_dcmdata_INCLUDE_DIR DCMTK_dcmimage_INCLUDE_DIR DCMTK_dcmimgle_INCLUDE_DIR DCMTK_dcmjpeg_INCLUDE_DIR DCMTK_dcmjpls_INCLUDE_DIR DCMTK_dcmnet_INCLUDE_DIR DCMTK_dcmpstat_INCLUDE_DIR DCMTK_dcmqrdb_INCLUDE_DIR DCMTK_dcmsign_INCLUDE_DIR DCMTK_dcmsr_INCLUDE_DIR DCMTK_dcmtls_INCLUDE_DIR DCMTK_ofstd_INCLUDE_DIR DCMTK_oflog_INCLUDE_DIR DCMTK_LIBRARIES) 
```

HelliceSaouli:
any luck i have the same problem here

zeripath:
The below appear to get rid of the initial problem:

```patch
From 96c9c8c02d8409bc955c7b0c5f2ea51938075971 Mon Sep 17 00:00:00 2001
From: Andrew Thornton <art27@cantab.net>
Date: Mon, 3 Aug 2020 17:48:46 +0100
Subject: [PATCH] Properly import dcmtk

Signed-off-by: Andrew Thornton <art27@cantab.net>
---
 core/vil/file_formats/vil_dicom.cxx      | 14 +++++++-------
[...]

PR #1045, opened 2026-05-11
BUILD: Drop bundled v3p/dcmtk and modernize vil_dicom for DCMTK 3.6+

### LINKED ISSUE #550: Does not build with new dcmtk:  API changed
In [this commit from 2009](https://git.dcmtk.org/?p=dcmtk.git;a=commit;h=15f4cd3c840f4e307d8037d05ec5d27eaad90b10), the declaration of the `DiInputPixelTemplate` function was changed. So, unless one uses the even older bundled version of dcmtk that vxl provides, one cannot build vxl with a native dcmtk installation that is up to date.

The current version of dcmtk is: [3.6.3 that was released this year](https://dcmtk.org/dcmtk.php.en). The version vxl bundles appears to be from 2002.

hjmjohnson:
Drop the bundled `v3p/dcmtk` (236 files, ~94k LOC) and modernize `vil_dicom` to read DICOM through an externally-installed DCMTK ≥ 3.6.5 via `find_package(DCMTK CONFIG)`. Closes #550, #768, #769, #816 by retiring the 23-year-old vendored fork and porting the reader to the modern public API.

<details>
<summary>Motivation</summary>

The bundled `v3p/dcmtk` dates to 2002 and has not built against any DCMTK ≥ 3.6 for roughly a decade. The relevant issues (#550, #768, #769, #816) were filed by distro/packaging maintainers (Fedora, FreeBSD ports, vcpkg) reporting that `vil_dicom` cannot be built against system DCMTK because of structural API changes (e.g. `DcmInputStreamFactory::ident()` becoming pure-virtual, `DiInputPixelTemplate`'s ctor growing from 6 to 9 parameters). No PR has addressed those reports in the seven years since the first one.

This PR makes the modernized reader build and run cleanly against DCMTK 3.7.0 and drops the vendored copy entirely.

</details>

<details>
<summary>What changed</summary>

Five focused commits:

| Commit | Subject |
|---|---|
| `2c74c205` | `BUILD:` Remove vendored `v3p/dcmtk` (236 files, ~94k LOC) |
| `46dbe80c` | `BUILD:` Use `find_package(DCMTK CONFIG)` for external DCMTK |
| `ab4be2fb` | `BUG:` Port `vil_dicom` to the DCMTK 3.6+ public API (fixes #550, #769) |
| `c6014eaf` | `BUG:` Read via `DcmFileFormat::loadFile()` to dodge static-lib ABI skew |
| `be7dee62` | `ENH:` Add minimal direct `vil_dicom` smoke test |

Highlights:

- `config/cmake/Modules/FindDCMTK.cmake` is replaced by `find_package(DCMTK CONFIG REQUIRED)`. `core/vil` links against the imported targets `DCMTK::dcmimgle`, `DCMTK::dcmdata`, `DCMTK::ofstd`.
- All nine DCMTK headers in the reader move to the namespaced form (`<dcmtk/dcmdata/dcfilefo.h>`, etc.).
- The 9-argument `DiInputPixelTemplate` ctor (`DiDocument*`, `..., DcmFileCache*, fragment&`) is used; vil's reader constructs a `DiDocument` over the loaded `DcmFileFormat` and passes a stack `DcmFileCache` with a zero fragment index.
- `VXL_BUILD_DCMTK` (which was only ever set by the deleted v3p subdir) is gone; `contrib/mul/vimt`'s `_BUILD_DCMTK` guard now keys on `VXL_USE_DCMTK`.

</details>

<details>
<summary>Why the input stream is no longer subclassed (commit c6014eaf)</summary>

The first port (`ab4be2fb`) used the historical `vil_dicom_stream_input` subclass of `DcmInputStream`. That compiled cleanly but reliably segfaulted on the very first virtual dispatch from `DcmFileFormat::readUntilTag`. The vtable pointer of the input-stream object read as the binary image base instead of the expected `DcmInputStream` vtable.

Root cause: DCMTK's static libraries are built with `-fvisibility=hidden` plus a private set of POSIX feature-test macros (`_DARWIN_C_SOURCE`, `_XOPEN_SOURCE_EXTENDED`, `_BSD_SOURCE`, `_BSD_COMPAT`, `_OSF_SOURCE`) that DCMTK does **not** propagate via `INTERFACE_COMPILE_DEFINITIONS` on its imported targets. A consumer that derives from `DcmInputStream` / `DcmProducer` therefore sees a slightly different class layout than the one frozen in `libdcmdata.a`, and the resulting ABI skew corrupts the vtable on the first virtual dispatch.

Trying to mirror DCMTK's internal flags from the consumer side (matching `CMAKE_CXX_STANDARD`, adding the POSIX macros, enabling `-fvisibility=hidden -fvisibility-inlines-hidden` on the affected sources) did not reliably fix the crash and is undocumented / platform-specific.

The robust fix is to **keep every DCMTK-derived object inside `libdcmdata.a`**: spool the `vil_stream` buffer to a `mkstemp()` temp file, hand the path to `DcmFileFormat::loadFile()`, and `unlink()` immediately. `loadFile()` uses DCMTK's own `DcmInputFileStream` entirely inside the library — no DcmInputStream-derived types remain on the consumer side, eliminating the ABI hazard.

The `vil_dicom_stream_{factory,producer,input}` classes and their header are deleted along with this change.

</details>

<details>
<summary>Testing</summary>

vxl ships six tiny DICOM fixtures in `core/vil/tests/file_read_data/` (sizes 988–2028 bytes) plus matching baselines that have been there for years. `test_file_format_read.cxx` already contained a DICOM test stanza gated on `HAS_DCMTK`; it just had never been exercised because no vxl build was wired against DCMTK in modern memory. With this PR both that stanza and a new dedicated `vil_test_dicom` smoke test pass.

Build configuration used locally:

```bash
cmake -G Ninja $SRC \
  -DVXL_BUILD_CORE_IMAGING=ON \
  -DVXL_USE_DCMTK=ON \
  -DDCMTK_DIR=/path/to/dcmtk-install \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_TESTING=ON
ninja vil_test_all
ctest -R "vil_test_(file_format_read|dicom)" -V
```

Results against DCMTK 3.7.0 on macOS arm64 (Apple Clang, `-O3`, `-std=c++17`):

```
1/2 Test #197: vil_test_file_format_read ........   Passed    0.75 sec
2/2 Test #198: vil_test_dicom ...................   Passed    0.01 sec
100% tests passed, 0 tests failed out of 2
```

The 6 DICOM cases inside `vil_test_file_format_read` (16/8-bit greyscale uncompressed variants and 12-bit float uncompressed) all pass. The new `vil_test_dicom` asserts 7 invariants (fixture opens, `make_input_image` returns a resource, format is `VIL_PIXEL_FORMAT_UINT_16`, single plane, non-zero dimensions, view decodes).

</details>

<details>
<summary>Compatibility</summary>

- `VXL_USE_DCMTK` is OFF by default, matching the existing surface area. Builds that did not enable it are unaffected.
- Builds that opted into `VXL_USE_DCMTK` previously expected the legacy `DCMTK_INCLUDE_DIR` / `DCMTK_LIBRARIES` module-mode variables. They now point `DCMTK_DIR` at an installed `DCMTKConfig.cmake` (or put DCMTK on `CMAKE_PREFIX_PATH`), matching every modern CMake package.
- Builds that previously relied on `VXL_FORCE_V3P_DCMTK` to use the bundled copy will fail to configure with a clear `find_package` error. Those builds should install a real DCMTK ≥ 3.6.5 or disable `VXL_USE_DCMTK`.
- The four packaging-maintainer issues (#550, #768, #769, #816) can be closed when this lands.

</details>


hjmjohnson:
@TimCootes This will need testing for your use cases.  I just updated to the best of the 23-year-old internal tests.

TimCootes:
Thanks for preparing this, Hans, that's really helpful.  I, or a colleague, will try it out soon.

hjmjohnson:
@TimCootes Did this update work for you?

TimCootes:
It works for some images but not for others.  My colleague is currently trying to find out why.

james-mcm-manchester:
@hjmjohnson Thanks for all the updates, some compressed dicoms failed to properly load into memory or decode. I've created a PR on your fork with fixes for those. (https://github.com/hjmjohnson/vxl/pull/1)

hjmjohnson:
Thanks @james-mcm-manchester — incorporated your encapsulated-DICOM fix in 70c6c8f (your authorship preserved; only your own added `MustRead` line was re-wrapped for clang-format).

Verified locally against a Homebrew-backed DCMTK build (DCMTK 3.7.0 from `/opt/homebrew`, which provides the `dcmjpeg`/`dcmjpls` components your fix links):

- `vil` builds and links cleanly with `DCMTK::dcmjpeg DCMTK::dcmjpls` added.
- `vil_test_file_format_read` — 84/84 passed
- `vil_test_dicom` — 7/7 passed

Note: the in-tree suite has no compressed/encapsulated DICOM fixture, so this confirms the new codec path compiles, links, and doesn't regress the uncompressed reads — it doesn't exercise actual JPEG/JPEG-LS decode. If you can contribute a small encapsulated fixture, I'm happy to wire up a decode test.

hjmjohnson:
@james-mcm-manchester. I did an AI code review and fixed a few exposed concerns.  Will you please verify that this new code base continues to work for you?

james-mcm-manchester:
@hjmjohnson Thanks very much for those, new codebase builds, tests and runs correctly.

PR #1046, opened 2026-06-08
ENH: Use portable std::lrint/floor/ceil for vnl_math rounding

hjmjohnson:
Replace the hand-written SSE2 / x87-asm / scalar rounding paths in `vnl_math` with portable `std::lrint`/`std::floor`/`std::ceil`. Modern compilers lower these to the same hardware rounding instruction the intrinsics emitted (`cvtsd2si` on x86-64) and to a native instruction on AArch64 (`fcvtns`), so the result is identical and fast on every architecture — and it speeds up the non-x86 path, which previously took the slow scalar fallback. `core/vnl/vnl_math.h`: **+20 / −349**.

<details>
<summary>What changed</summary>

`rnd` and `rnd_halfinttoeven` use `std::lrint` (round-to-nearest, default ties-to-even); `floor`/`ceil` use `std::floor`/`std::ceil`; `rnd_halfintup` keeps the `2*x + 0.5` half-to-even trick (no standard ties-up primitive). Removed:

- the `<emmintrin.h>` include,
- the `USE_SSE2_IMPL` / `GCC_USE_FAST_IMPL` / `VC_USE_FAST_IMPL` macros and the four-way `#if` split (SSE2 intrinsics, GCC `fistpl` inline asm, MSVC `__asm`, scalar vanilla),
- the now-unused `VNL_CHECK_FPU_ROUNDING_MODE` asserts that guarded the fast paths.

`VNL_CONFIG_ENABLE_SSE2_ROUNDING` / `VXL_HAS_EMMINTRIN_H` are no longer referenced by `vnl_math.h`; they remain defined by the build and can be retired in a follow-up if desired.
</details>

<details>
<summary>Validation</summary>

- **Equivalence (x86-64, gcc 13.3, -O2):** a differential test comparing the old SSE2 path against the new `std::` path over a dense sweep — 8,000,001 doubles across [−1e6, 1e6] at step 0.25 (every tie and quarter) plus 800,001 floats — reports **0 mismatches**. Disassembly confirms the new code lowers to `cvtsd2si`.
- **Full build with contrib** (`VXL_BUILD_CONTRIB=ON`, `BUILD_TESTING=ON`): **4370/4370 targets, 0 errors**.
- **Full test suite** (core + all contrib): **705 / 705 passed**.
- `vnl/tests/test_math` (120 rounding assertions: ties-to-even, ties-up, near-half, near-integer, ±, even/odd, float and double) passes on x86-64 and arm64.

The two intentional out-of-range differences are improvements: for `|x| > INT_MAX/2` the old `2*x` trick overflowed (UB) while the new `floor`/`ceil` are correct; NaN/Inf are degenerate.

Conclusion: old ≡ new for every finite value in the operating range, validated across the full contrib test matrix on x86-64.
</details>

### DIFF
--- a/core/vnl/vnl_math.h
+++ b/core/vnl/vnl_math.h
@@ -45,34 +45,6 @@
 #include <vxl_config.h>
 #include <vnl/vnl_config.h> // for VNL_CONFIG_ENABLE_SSE2_ROUNDING
 #include <vnl/vnl_export.h>
-#ifdef VNL_CHECK_FPU_ROUNDING_MODE
-#  include <cassert>
-#endif
-
-// Figure out when the fast implementation can be used
-#if VNL_CONFIG_ENABLE_SSE2_ROUNDING && defined(__SSE2__)
-#  if !VXL_HAS_EMMINTRIN_H
-#    error "Required file emmintrin.h for SSE2 not found"
-#  else
-#    include <emmintrin.h> // sse 2 intrinsics
-#    define USE_SSE2_IMPL 1
-#  endif
-#else
-#  define USE_SSE2_IMPL 0
-#endif
-
-// Turn on fast impl when using GCC on Intel-based machines with the following exception:
-#if defined(__GNUC__) && ((defined(__i386__) || defined(__i386) || defined(__x86_64__) || defined(__x86_64)))
-#  define GCC_USE_FAST_IMPL 1
-#else
-#  define GCC_USE_FAST_IMPL 0
-#endif
-// Turn on fast impl when using msvc on 32 bits windows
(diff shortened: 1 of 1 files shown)


hjmjohnson:
Moving ITK efforts for VXL to https://github.com/InsightSoftwareConsortium/vxl/tree/for/itk-vxl-master.  This is primary so that the ITK development team can focus on the ITK-needed features of vxl and remove the rest.

decrispell:
@hjmjohnson Any reason to close this, other than moving your efforts to the ITK vxl fork?  It looks like a nice cleanup to me.

PR #1047, opened 2026-06-08
BUILD: Export vxl_platform_math even when VXL_NO_EXPORT is ON

hjmjohnson:
Export the `vxl_platform_math` INTERFACE target even when `VXL_NO_EXPORT` is ON, so that VXL can be embedded in a host project (ITK) that exports its own targets.

`config/cmake/Modules/ConfigurePlatformMathTarget.cmake` only added `vxl_platform_math` to the install export set under `if(NOT VXL_NO_EXPORT)`. But `VXL_NO_EXPORT=ON` is exactly the embedding mode: `itkvcl` and `itkv3p_netlib` link `vxl_platform_math` and are placed in the host's export set, so CMake requires the interface target to be in an export set too. With the guard, `install(EXPORT ...)` fails at generate time:

```
install(EXPORT "ITKTargets" ...) includes target "itkvcl" which requires
target "vxl_platform_math" that is not in any export set.
```

The fix removes the guard so the interface target is always exported. It has no artifacts, so this is a no-op for standalone builds.

<details>
<summary>Validation</summary>

- Standalone `cmake` generate succeeds with the change, both default and `-DVXL_NO_EXPORT=ON`.
- This is the same export-set arrangement already used successfully when this tree is embedded in ITK.
</details>

### DIFF
--- a/config/cmake/Modules/ConfigurePlatformMathTarget.cmake
+++ b/config/cmake/Modules/ConfigurePlatformMathTarget.cmake
@@ -2,14 +2,14 @@
 add_library(vxl_platform_math INTERFACE)
 # Ensure the interface target participates in the export set so that
 # exported libraries depending on it do not trigger CMake export errors.
-# INTERFACE libraries have no artifacts, but can and must be exported
-# if referenced by other exported targets.
-if(NOT VXL_NO_EXPORT)
-    # Participate in build-tree export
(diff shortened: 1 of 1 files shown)


hjmjohnson:

dzenanz (approved):

PR #1048, opened 2026-06-08
BUILD: Require C++17 in VALID_CXX_STANDARDS

hjmjohnson:
Drop `"11"` and `"14"` from `VALID_CXX_STANDARDS` so that VXL 7's C++17 minimum is actually enforced.

VXL 7 defaults `CMAKE_CXX_STANDARD` to `17` and uses C++17-only constructs (`inline constexpr` variables, structured bindings), but the validity list still accepted `11`/`14`. A build configured with an explicit `-DCMAKE_CXX_STANDARD=11` passed the check and then failed to compile. With this change such a configuration is rejected up front with a clear message.

<details>
<summary>Validation</summary>

- Default configure (`CMAKE_CXX_STANDARD=17`): succeeds.
- `-DCMAKE_CXX_STANDARD=11`: now fails at configure time with
  `CMAKE_CXX_STANDARD:STRING=11 is not a supported C++ standard; VXL requires one of: 17;20;23.`
</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,9 +29,12 @@ endif()
 if(NOT CMAKE_CXX_EXTENSIONS)
   set(CMAKE_CXX_EXTENSIONS OFF)
 endif()
-set(VALID_CXX_STANDARDS "11" "14" "17" "20" "23")
+# VXL 7 requires C++17 or newer (inline constexpr variables, structured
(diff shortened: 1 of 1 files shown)


hjmjohnson:

dzenanz (approved):

PR #1049, opened 2026-06-08
COMP: Drop redundant int cast in vnl_rnpoly_solve powr() calls

hjmjohnson:
Remove the redundant `static_cast<int>` in the two `powr()` calls in `vnl_rnpoly_solve.cxx::initr`.

`powr()` takes an `unsigned int` and `ideg` is `std::vector<unsigned int>`, so `static_cast<int>(ideg[j])` casts the value *down* to `int` only for it to implicitly widen back to `unsigned int` at the call — an unnecessary sign-conversion round-trip. Passing `ideg[j]` directly is the exact-type match.

<details>
<summary>Validation</summary>

`vnl_rnpoly_solve.cxx` compiles cleanly with the change (no sign-conversion warning).
</details>


hjmjohnson:

dzenanz (approved):

PR #1050, opened 2026-06-08
COMP: Validate size before narrowing in vnl_powell_1dfun delegating ctor

hjmjohnson:
Validate the size before narrowing `size_t` to `int` in the delegating `vnl_powell_1dfun` constructor, so the bounds check runs *before* the conversion.

The `size_t`-taking constructor delegated with `static_cast<int>(n)` in its member-initializer list, which is evaluated before the constructor body. The `assert(n <= INT_MAX)` therefore ran only after the narrowing had already happened: for `n > INT_MAX` the delegated `int` constructor was invoked with a wrapped value before the guard could fire. The fix moves the check into a small helper (`vnl_powell_checked_int`) that asserts and then converts, called from the initializer list.

<details>
<summary>Validation</summary>

`vnl_powell.cxx` compiles cleanly with the change.
</details>

### DIFF
--- a/core/vnl/algo/vnl_powell.cxx
+++ b/core/vnl/algo/vnl_powell.cxx
@@ -24,6 +24,20 @@
 #  endif
 #endif
 
+namespace
+{
+//: Validate that a size_t fits in an int before narrowing, so the guard
+// runs before the conversion is used. A delegating constructor evaluates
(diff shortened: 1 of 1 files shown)


hjmjohnson:

dzenanz (approved):

PR #1053, opened 2026-08-27
COMP: Use an absolute path in the core tests EXISTS guard

hjmjohnson:
`if(EXISTS)` is well defined only for absolute paths, so the `core/tests` guard resolved against the **cmake process working directory** instead of the current source directory. In an ordinary build `core/tests` is skipped and three tests never run.

Configuring with `BUILD_TESTING=ON` from a neutral directory:

| | tests |
|---|---|
| `master` (1450e99) | 428 |
| this branch | **431** |

The three recovered: `cmake_config`, `core_test_build_info`, `core_test_config`.

No longer blocked: #1058 has merged, so the invalid `#ifdef` that compiling `core/tests` exposed is already fixed on master. This branch is rebased on top of it and builds clean.

<details>
<summary>Reproducing both directions</summary>

The guard's own comment says it exists so consumers that strip the directory don't break (`# NOTE: some external projects remove the tests directory (aka ITK)`). It fails at that too, in the opposite direction: if the working directory happens to contain a `tests` subdirectory, the guard passes and `add_subdirectory` fails on a directory that isn't there.

```
CMake Error at core/CMakeLists.txt:33 (add_subdirectory):
  add_subdirectory given source "tests" which is not an existing directory.
```

Minimal demonstration, CMake 3.26.6:

```
cmake run from a directory containing ./tests  ->  EXISTS tests -> TRUE
cmake run from a directory without   ./tests  ->  EXISTS tests -> FALSE
```

So the same line both skips vxl's own tests and breaks strip-the-directory consumers, depending only on where cmake was invoked from.

</details>

<details>

macOS 15 arm64, CMake 3.26.6, Ninja, same working directory and flags for both:

```bash
cmake -S <vxl> -B <build> -GNinja -DBUILD_TESTING=ON
ctest --test-dir <build> -N
```

`core/tests` is absent from the build tree before and present after, gaining 7 build targets. Configure succeeds in both cases — the failure is silent, which is why it has gone unnoticed.

</details>

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -142,7 +142,7 @@ add_subdirectory(testlib)
 
 # Tests that check and output the vxl configuration
 # NOTE: some external projects remove the tests directory (aka ITK)
-if( BUILD_TESTING AND EXISTS "tests")
+if( BUILD_TESTING AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/tests")
   add_subdirectory(tests)
 endif()
 if( VXL_BUILD_EXAMPLES )
--- a/core/tests/test_config.cxx
+++ b/core/tests/test_config.cxx
@@ -25,14 +25,6 @@ test_config()
 
   std::cout << "All macro definitions are shown inside parentheses\n";
 
-  std::cout << "template <> ";
-#ifdef template < >
-  std::cout << "is set to " << quote(template <>);
-#else
-  std::cout << "is not set";
-#endif
-  std::cout << std::endl;
-
   std::cout << "VXL_LITTLE_ENDIAN ";
 #ifdef VXL_LITTLE_ENDIAN
   std::cout << "is set to " << quote(VXL_LITTLE_ENDIAN);
--- a/core/vil/file_formats/vil_nitf2_field_functor.h
+++ b/core/vil/file_formats/vil_nitf2_field_functor.h
@@ -152,9 +152,10 @@ private:
 class vil_nitf2_max_field_value_plus_offset_and_threshold : public vil_nitf2_field_functor<int>
 {
 public:
-  vil_nitf2_max_field_value_plus_offset_and_threshold(
-    std::string tag, int offset, int min_threshold = 0, int tag_factor = 1
-  )
+  vil_nitf2_max_field_value_plus_offset_and_threshold(std::string tag,
+                                                      int offset,
+                                                      int min_threshold = 0,
+                                                      int tag_factor = 1)
     : tag_(std::move(tag))
     , offset_(offset)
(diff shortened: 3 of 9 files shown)


decrispell:
Thanks @hjmjohnson !  Did you mean to include all the formatting changes with this PR?

hjmjohnson:
@copilot resolve the merge conflicts on this branch.  Remove the formatting changes

hjmjohnson:
Good catch — no, the formatting changes did not belong here. Dropped, and the branch is rebased on current master.

Two force-pushes, one concern each: `6d6d105..5bb3fee` removes the formatting commits (content only, old base), `5bb3fee..43dd429` is the plain rebase onto master.

The PR is now two commits touching two files: `core/CMakeLists.txt` (the absolute-path guard) and `core/tests/test_config.cxx` (an invalid `#ifdef` the guard fix exposes).

<details>
<summary>Why the formatting commits were there, and why they are not needed now</summary>

They were added only to get `pre-commit run --all-files` to pass on a tree that already had violations on master — trailing whitespace in `core/vil/file_formats/vil_tiff_header.cxx` and six files needing a clang-format re-wrap. That was the wrong call: it is unrelated to this fix and it is what made the branch conflict.

Correction to an earlier version of this comment: I first read master as having picked up the clang-format fixes on its own. It has not. `pre-commit` runs `trailing-whitespace` before `clang-format` and aborts as soon as a hook modifies a file, so `--all-files` stops at the whitespace line and never reports what follows. Both are still pre-existing on master, and neither is touched by this PR.

The whitespace line is now #1057. The six clang-format files behind it are still outstanding; say the word and they can follow as their own `STYLE:` PR.

</details>

<details>
<summary>Verification on the rebased branch</summary>

Configured as the `itk-core` job does (`VXL_BUILD_CORE_NUMERICS_ONLY=ON`, `BUILD_TESTING=ON`, Release):

```
ninja   -> 0 warnings
ctest   -> 100% tests passed, 0 tests failed out of 116
```

`pre-commit run --files core/CMakeLists.txt core/tests/test_config.cxx` passes every hook, clang-format included.

</details>

hjmjohnson:
Split so each commit is its own PR. This one is now the `core/CMakeLists.txt` guard fix alone — one commit, one file.

Merge order matters: **#1058 first, then this one.** The guard makes `core/tests` compile for the first time, which exposes a pre-existing invalid `#ifdef` in `test_config.cxx`; with `--fail-on-warnings` that is fatal, so CI here will be red until #1058 merges. Measured on this branch: exactly 1 warning, and it is that one.

`#1057` (the `vil_tiff_header` whitespace line) is independent of both and can go in any order.

